PR #10, opened 2023-06-01
Integrator api initial version

liviuba:
Will probably be changed as we start using it for the client, but the general structure should be mostly fixed.

See Readme.md "Running the api" for how to run


kbab (approved):
Since these are API related, I assume there are no conflicts with things in the main branch

PR #13, opened 2023-06-19
Display the correct license on generated image pages

matthewh-ebi:
Currently we read the license from the original BioStudies entry on ingest, which is part of our data model. However, when we render the license on generated image pages, we always show CC0, which is incorrect. This PR fixes this for the two licenses we allow for web submission.

### DIFF
--- a/static-sitegen/scripts/generate_image_page.py
+++ b/static-sitegen/scripts/generate_image_page.py
@@ -14,6 +14,11 @@ from utils import format_for_html
 logger = logging.getLogger(os.path.basename(__file__))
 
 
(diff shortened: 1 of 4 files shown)


AybukeKY (commented):
There are 2 commits in this code: first one related to mapping channel and z-plane structure of images and converting them, second one on displaying the correct licence uri.
Do you want me to review first one as well?

On licence: In the models, the default value of licence is CC0. Do we want to default to Unknown? Because in this code, license_uri default is Unknown (which it will never be as licence defaults to CC0 anyway).

PR #14, opened 2023-06-19
Biaint api backend

liviuba:


clients/python/.github/workflows/python.yml
```diff
@@ -0,0 +1,37 @@
...
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+      - name: Lint with flake8
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+      - name: Test with pytest
+        run: |
+          pytest
```
ctr26, line 37: I try to put my test commands in the Makefile and the call 

  make test

from github actions.

That way if we change CI provider it'll be a little easier to completely decouple

ctr26:
@liviuba it looks like this needs rebasing on main

ctr26:
@liviuba the poetry guidelines suggest commititng poetry lock rather than ignoring. I use a precommit to keep poetry lock synced for instance

liviuba:
> @liviuba the poetry guidelines suggest commititng poetry lock rather than ignoring. I use a precommit to keep poetry lock synced for instance

It was a quick fix for a problem I was having deploying to dev.

Can you PR the precommit hooks to this branch, please?

ctr26:
I don't supposed there's anyway of breaking this PR into smaller ones?

liviuba:
> I don't supposed there's anyway of breaking this PR into smaller ones?

I'd like to, but I don't see how. Most things downstream depend on how we represent/store the data, and the api itself comes as one big chunk for the initial release.

IMO the least painful way to do the review is to pair (/multi-pair if anyone else wants to get involved?) for a few walkthroughs (something like 1h chunks until we go over it entirely) to speed things up, wait for a week to let it sink in / surface less obvious issues, merge and 🤞

api/src/api/public.py
```diff
@@ -199,6 +199,57 @@ async def search_file_references_exact_match(
...
+async def search_studies_exact_match(
+        annotations_any: Annotated[List[api_models.SearchAnnotation], Body()] = [],
+        study_match: Annotated[Optional[api_models.SearchStudy], Body()] = None,
+        start_uuid: Annotated[Optional[UUID], Body()] = None,
+        limit : Annotated[int, Body(gt=0)] = 10,
+        db: Repository = Depends()
+    ) -> List[db_models.BIAStudy]:
+    query = {}
+    if annotations_any:
+        obj_annotations_any = [query_term.model_dump(exclude_defaults=True) for query_term in annotations_any]
```
ctr26, line 212: Need to run this through black formatter

matthewh-ebi (approved):
Looks good to me. I will be celebrating this merge (though probably just with coffee).

PR #22, opened 2023-06-26
Add/update 2D plane rendering code

matthewh-ebi:
This PR adds new rendering code, replacing the current thumbnail generation.

It currently only supports creating 2D images from OME-NGFF image representations by selection specific T/Z values (i.e. only XY plane rendering works, XZ or ZY plane rendering does not).

The underlying functions support selection of a 2D bounding box within the plane, windowed channels, and selection of channel colours. However these functions can only be invoked manually, the thumbnail generation code guesses default values for all settings.

The code makes a number of shortcuts (for example, it cannot handle NGFF images that are not either TCZYX or ZYX), with the goal of fixing 2D rendering first and expanding later.


kbab (approved):
Everything looked good to me. Only one minor question about placement of DEFAULT_BB on line 260 of rendering.py. I guess it is because of context ...

tools/bia_integrator_tools/rendering.py
```diff
@@ -0,0 +1,352 @@
...
+    
+    1. Lazy-load the image as a Dask array.
+    2. Select the plane (single t and z values) we'll use.
+    3. Separate channels.
+    4. Apply a color map to each channel array.
+    5. Merge the channel arrays."""
+
+    ydim, xdim = dims
+
+    min_ydim_needed = ydim / bbrel.ysize
```
matthewh-ebi, line 274: Comment this

PR #23, opened 2023-06-27
Dev

ctr26:
@kbab 

Here is an attempt at helping with your snakemake "resumability" problem. 

You add a touch(file) flag output that snakemake will generate when the script doesn't crash, you then use that as an input to the next rule

https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#flag-files

P.s. I've not tested this feel free to commit to this branch


ctr26:
@kbab Did this help? Can I close/merge?

kbab:
@ctr26 Yes it did - it was very helpful! Apologies for not replying sooner. You can close, no need to merge, I included it already. Thanks!

PR #24, opened 2023-06-29
Add filetype info (and other changes for pipelining conversion to zarr)

kbab:
The main theme of this PR is obtaining filetype information from studies.

Most of the commits in this PR relate to obtaining filetype information from studies and displaying them in the static html pages.

However, there are some other commits that addressed issues experienced when running the snakemake pipeline on 
studies submitted between Jan2023 and May2023.


static-sitegen/scripts/generate_dataset_page.py
```diff
@@ -24,6 +24,15 @@
 
 DEFAULT_TEMPLATE = "dataset-landing.html.j2"
 
+# Attributes to display in study content section
+# Use list for backward compatibility with different annotation names
+# Most recent names come first in list
+STUDY_CONTENT_ANNOTATIONS = {
+    "Study size:": ["study_size_human_readable", "study_size"],
```
liviuba, line 31: If this is an attribute of the study, could it be study_size_bytes, and then prettified when displaying? For normalization and avoiding confusion (e.g. to make it a number with implicit units, instead of mb/Mb/Mib etc)

tools/resources/readme.txt
```diff
@@ -0,0 +1,6 @@
+Notes on creating the files in this directory:
+
+1. Copy supported formats from https://bio-formats.readthedocs.io/en/stable/supported-formats.html
+2. Paste into spreadsheet or text editor and get all extensions
```
liviuba, line 4: Copy-paste friendly alternative in the developer console, with the supported-formats page opened:

```js
[...new Set(Array.from(document.getElementsByTagName("tbody")[0].querySelectorAll("td:nth-child(2)")).map(el => el.innerText.split(",")).flat().filter(extension => extension.length))].sort()
```

liviuba (approved):

static-sitegen/scripts/generate_dataset_page.py
kbab: Interesting point. I think @matthewh-ebi has mentioned something similar in the past too. I have created a github issue  #26  to discuss refactoring - I hope this is ok for the PR.

Issue #31, opened 2023-07-24
Assign types to zipped filerefs when ingesting from biostudies


kbab:
We currently do not use the 'type' field in the FileReference object when ingesting from biostudies (although we use it when indexing from zips and possibly in other places). Running the Snakemake pipeline on the S-BIAD687 and some other zipped studies presented an issue causing the pipeline to fail because some directories were zipped when loaded into FIRE, but their names did not end with '.zip'.

In tools/scripts/generate_image_download_size_annotations.py, the sizes of zipped archives are stored in a lookup dict. However, only filerefs  whose names end with '.zip' are inserted into this dict. So directories zipped and put on FIRE are not identified.

To address this, the intention is to use the 'type' field of the Fileref object to store if a file is a zipped archive. Criteria for this is

1. If name ends with .zarr.zip -> type=zipped_zarr (see #27)
2. If name ends with .zip -> type=zipped_archive
3. If name does not end with a recognised extension but there is an endpoint when '.zip' added to uri -> type=zipped_dir

kbab:
After discussions within the team we decided to use the 'type' attribute returned by the biostudies api which classifies file references as either 'file' or 'directory'

PR #32, opened 2023-07-27
Snakemake keepgoing

kbab:
This PR stems from #31 and includes downstream modifications to use the 'FileReference.type' attribute for zips

The main change is the assignment of one of 'zipped_zarr', 'zipped_archive' or 'zipped_directory' to file references on ingestion from biostudies. Some downstream scripts were modified to make use of the zip types.

Tests and mock data were added to test that ingest_from_biostudies was making the zip assignments as expected.


tools/scripts/ingest_from_biostudies.py
```diff
@@ -24,6 +25,13 @@
 
 logger = logging.getLogger(__file__)
 
+# ToDo: Discuss whether to put this somewhere common so it can be reused
```
matthewh-ebi, line 28: It should be common I think, and abstracted out so that if we end up using a more reliable method for determining file type (like magic bytes) we don't have to change code everywhere.

tools/scripts/ingest_from_biostudies.py
```diff
@@ -39,6 +47,23 @@ def bst_file_to_file_reference(accession_id: str, bst_file: File, file_uri_templ
         attributes=fileref_attributes
     )
 
+    if fileref.name.endswith(".zarr.zip") or fileref.name.endswith(".zarr"):
```
matthewh-ebi, line 50: Discussion point for review: I think at this stage, we shouldn't be distinguishing between zipped zarrs and other archive files. We can do that when we come to image assignment. We do need to get the URIs right for folders though.

tools/scripts/ingest_from_biostudies.py
```diff
@@ -100,7 +125,17 @@ def study_links_from_submission(submission: Submission) -> dict:
     return links
 
 
-def imaging_method_from_submission(submission: Submission) -> str:
+def _get_with_case_insensitive_key(dictionary: dict, key: str) -> str:
```
matthewh-ebi, line 128: Generic code like this should move to biostudies.py, this is where the existing code for searching (e.g. for File Lists) is.

tools/scripts/ingest_from_biostudies.py
```diff
@@ -100,7 +125,17 @@ def study_links_from_submission(submission: Submission) -> dict:
...
+    keys = [k.lower() for k in dictionary.keys()]
+    temp_key = key.lower()
+    if temp_key in keys:
+        key_index = keys.index(temp_key)
+        temp_key = list(dictionary.keys())[key_index]
+        return dictionary[temp_key]
+    else:
+        raise KeyError(f"{key} not in {dictionary.keys()}")
+
+def get_from_subsection(submission: Submission, key: str) -> str:
```
matthewh-ebi, line 138: This is called get_from_subsection, but the call argument is a Submission object, which is a bit confusing. Docstring would be useful here.

tools/scripts/ingest_from_biostudies.py
```diff
@@ -109,12 +144,34 @@ def imaging_method_from_submission(submission: Submission) -> str:
...
+    
+    # Check other subsections
+    other_subsections = [
+        section
+        for section in submission.section.subsections
+        if section.type != 'Study Component'
+    ]
 
-    return imaging_method
+    if len(other_subsections):
```
liviuba, line 164: Nitpick, but this is a bit confusing to me, since the for wouldn't do anything if other_subsections is empty

tools/tests/src/test_ingest_from_biostudies.py
```diff
@@ -0,0 +1,64 @@
...
+                self.content = file_to_get.read_text()
+            except Exception:
+                raise Exception(f"Could not read {file_to_get}. url={url}")
+        def json(self):
+            return self.json_data
+    return MockResponse()
+
+# Note below that two dummy variables (mock_head and mock_get) are passed
+# due to the two mock.patch decorators
+@mock.patch("bia_integrator_tools.biostudies.requests.get", side_effect=mocked_requests_get)
```
liviuba, line 30: 👍

liviuba (approved):

tools/scripts/ingest_from_biostudies.py
kbab: Moved as suggested

PR #46, opened 2023-08-14
Bia ninja2 template

kbab:
This branch allows giving a consistent look and feel to the visualisation pages by extension of a base template which frames contents around the current BIA website look. Google analytics were incorporated into the base template, and modifications to the collection, dataset and image pages were made:

1. If no zips in study display single line for number of files
2. Modify base template so content uses more screen real estate
3. Allow sorting of columns in collection landing page
4. Include release date in collection landing page
5. Make dataset pages more like those in AI collection
    - remove study type
    - add license
    - add link to original submission
6. Make image landing pages more web analytics friendly
    - change viewer buttons to links (so they are captured as events)
    - add accession number and image alias to title

### DIFF
--- a/static-sitegen/scripts/generate_dataset_page.py
+++ b/static-sitegen/scripts/generate_dataset_page.py
@@ -27,10 +27,16 @@ DEFAULT_TEMPLATE = "dataset-landing.html.j2"
 # Attributes to display in study content section
 # Use list for backward compatibility with different annotation names
 # Most recent names come first in list
-STUDY_CONTENT_ANNOTATIONS = {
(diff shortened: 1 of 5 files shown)


matthewh-ebi (approved):
This is great, and we should definitely merge. Many thanks, Kola. Couple of things to think about (no need to do anything about these now though):

a) We'll probably want to think a bit about positioning of the "Alpha" banner and the way the general BIA website and these pages fit together - we can do this with some later tweaks.

b) We definitely need to plan for how we keep 1) the base template here, 2) the BIA static webpages and 3) the BioStudies magic JS that mirrors the look of the BIA site in sync. This will be hard (BioStudies already doesn't show the Policies tab for example).

Issue #37, opened 2023-08-02
Recognise compound extensions when summarising filetypes


kbab:
Summarising filetypes currently uses just the suffix (string after last '.'). However, some files have compound extensions e.g. ome.tiff, zarr.zip and ome.zarr.zip. These need to be catered for.

Issue #38, opened 2023-08-03
Handle zarrs in snakemake workflow


kbab:
The current workflow has  a step that converts images to ome zarrs. However, we are now receiving submissions with zarrs (OME and non-ome) zarrs.

We need the conversion stage to be able to handle zarrs (what do we do about non OME zarrs?). Also related to issue #37 (recognising compound filetypes).

We can test on S-BIAD815. Also see details on https://bia-integrator.readthedocs.io/en/latest/staging_zipped_zarrs.html

Issue #27, opened 2023-07-04
Register zarrs differently to other file references


kbab:
The root of the zarr format is a directory. This is stored on FIRE as a zip archive usually with the suffix .zarr.zip. We need to create a workflow to deal with this format.
1. Need to copy from FIRE, unzip, then copy to S3
2. Annotate the fileref with type 'zarr' or 'zarr_archive'?
3. When summarising filetypes just get the size of the archive (do we need both compressed and uncompressed sizes)?

kbab:
![](https://github.trello.services/images/mini-trello-icon.png) [Register zarrs differently to other file references](https://trello.com/c/cixuSrXT/146-register-zarrs-differently-to-other-file-references)

matthewh-ebi:
There's a prototype of this in https://github.com/BioImage-Archive/bia-integrator/blob/main/tools/scripts/assign_zipped_zarr_from_filerefs.py, it uses "zipped_zarr" as the type, but we don't have to stick to that.

PR #47, opened 2023-08-14
Handle zarrs in snakemake workflow

kbab:
This branch allows the current snakemake workflow to handle zarr files using the existing remote_zipped_zarr_to_s3.py script. It currently does not distinguish between OME and non-ome zarrs. It simply downloads zarrs unzips if necessary and uploads to S3.

Fixes issues #37 and #38 and allows #27 to be closed.

### DIFF
--- a/tools/scripts/assign_zipped_zarr_from_filerefs.py
+++ b/tools/scripts/assign_zipped_zarr_from_filerefs.py
@@ -22,7 +22,7 @@ def main(accession_id):
     filerefs = bia_study.file_references
 
     for fileref in filerefs.values():
-        if "ome.zarr" in fileref.name:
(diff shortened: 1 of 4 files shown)


matthewh-ebi (commented):
Looks good.

One suggestion: it's being aware that assign_zipped_zarr_from_filerefs.py just looks to see whether `"ome.zarr"` is in the fileref name. So if that string isn't present, it won't get assigned as an image, and so I don't think it'll be staged to S3. We should test this and probably add a comment to to the snakemake file so we don't get caught out later.

AybukeKY (approved):
All good. I agree with Matthew on adding a comment on just handling ome.zarr extensions at the moment.

PR #50, opened 2023-08-17
Fix Download URL for zipped_zarrs in dataset and image pages

kbab:
The addition of the zipped zarrs to the snakemake workflow introduced a bug where the download urls for the zipped zarr files were not being added to the dataset and image pages during static_site page generation. This PR fixes it.

### DIFF
--- a/static-sitegen/scripts/generate_dataset_page.py
+++ b/static-sitegen/scripts/generate_dataset_page.py
@@ -9,7 +9,8 @@ from bia_integrator_core.integrator import load_and_annotate_study
 from bia_integrator_core.interface import get_aliases
 from utils import ( get_annotation_files_in_study, 
                    get_non_annotation_images_in_study,
-                   add_annotation_download_size_attributes
+                   add_annotation_download_size_attributes,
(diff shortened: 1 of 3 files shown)


static-sitegen/scripts/generate_dataset_page.py
```diff
@@ -81,7 +81,7 @@ def generate_dataset_page_html(accession_id, template_fname: str):
                 image_landing_uris[image.id] = f"{accession_id}/{image.id}.html"
             if representation.type == "thumbnail":
                 image_thumbnails[image.id] = representation.uri
-            if representation.type == "fire_object":
+            if representation.type in ["fire_object", "zipped_zarr"]:
```
matthewh-ebi, line 84: The representation types for which we provide download URIs should probably be a variable defined somewhere and imported by both `generate_dataset_page.py` and `generate_image_page.py`, otherwise if we end up adding new representation types, we'll probably forget to add them again.

matthewh-ebi (approved):
Looks good (though would recommend the suggested refactoring of representation types as a shared variable).

PR #67, opened 2023-09-25
Update VEM collection

matthewh-ebi:
This PR:

a) Updates the VEM collection to use the AI datasets collection landing style.
b) Changes neuroglancer link handling (only display if the link exists as an attribute, since it doesn't work well for many images).
c) Adds a deploy script to push to the newly provisioned staging bucket for pages (so we can test without breaking the existing pages).

New VEM pages visible here:

https://uk1s3.embassy.ebi.ac.uk/bia-pages-staging/pages/vem.html


static-sitegen/deploy-pages-staging.sh
kbab, line 1: This is a helpful utility!

AybukeKY (approved):
Looks all good. I assume _neuroglancer_link will change in the future?

PR #70, opened 2023-10-06
Api container dep

ctr26:
@liviuba Is there any reason to go via ghcr for the api image when we can just build it?

### DIFF
--- a/.github/workflows/api.yaml
+++ b/.github/workflows/api.yaml
@@ -29,10 +29,10 @@ jobs:
       - name: Docker Compose Up
         run: |
           docker-compose up --build -d  # Build and start your Docker containers in the background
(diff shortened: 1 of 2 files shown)


liviuba:
> @liviuba Is there any reason to go via ghcr for the api image when we can just build it?

Good question, I'm not sure, habit I guess. Thanks for the PR!

Can you please raise this PR for biaint-api-backend instead? That's where api work happens now that it isn't quite released yet, but things are migrating to use it so would break if on main.

ctr26:
I'd prefer merge this small PR and then rebase biaint-api-backend on main

liviuba:
Merging this to get the build on compose up bit

PR #72, opened 2023-10-07
Automatic building on api image

ctr26:
Automatic building on api image

### DIFF
--- a/.github/workflows/publish_container.yaml
+++ b/.github/workflows/publish_container.yaml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
       - master
-  pull_request: ~
+  # pull_request: ~
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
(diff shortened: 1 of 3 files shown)


ctr26:
Future issues are that this builds api images every main branch commit, which would be fine and sensible for a repo with only one job. Maybe a better future solution would be tag based? 

I'd also suggest that the docker compose api bits go in the root dir and we just have a master docker-compose for the static site and the api

I've tested this on my fork and it's working

ctr26:
/\ I say this because I had issues setting this up using working directories and makefiles not in root dir

ctr26:
@mrmh2 @liviuba thoughts?

matthewh-ebi:
Fine for me for now. I think we'll need to be much further on with k8s and deployment before we can practically use/test this though.

liviuba:
Looks good, the api doesn't actually pull the version from pyproject.toml, but it should (/will eventually)

For the unnecessary rebuilds, we might be able to only filter for specific directory changes when triggering actions to make them sub-project specific (this might be supported out of the box IIRC). Also, part of the reason for the monorepo was to make it easier to keep things in sync, so "release tags" for the entire repo would sort of make sense. Something like CI/CD for everything, but major versions get a tag?

I also feel like this is a problem for our future selves :D

matthewh-ebi (approved):

PR #74, opened 2023-12-07
Migrate images only

liviuba:
Add cli command to migrate images only, --env otion


data-migration/Readme.md
kbab, line 1: Looks OK. I noted you use os.path instead of the newer pathlib - not sure what official guidance from python on this is.

kbab (approved):
See comment about os.path

PR #75, opened 2023-12-08
This makes snakemake check for env variables and warn you if they're not set

ctr26:
I've specifically not added so env vars to this patch because they have sane defaults


kbab (approved):

matthewh-ebi:
Yes - looks fine to me

PR #76, opened 2023-12-08
Generate static page artefacts via api

kbab:
This is a rehash of #73 with cleaner commits. It allows generation of static pages using the API. However, it does not address the recent issue on the use of load_and_annotate_study (except in the static-sitegen/generate_image_page.py script).

### DIFF
--- a/static-sitegen/scripts/build_collection_site.py
+++ b/static-sitegen/scripts/build_collection_site.py
@@ -4,6 +4,7 @@ from pathlib import Path
 import click
 
 from bia_integrator_core.collection import get_collection
+from bia_integrator_core.study import get_study
 from bia_integrator_core.integrator import load_and_annotate_study
 
 from generate_collection_page import generate_collection_page_html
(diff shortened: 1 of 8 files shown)


static-sitegen/scripts/build_collection_site.py
```diff
@@ -42,12 +43,16 @@ def main(collection_name):
     output_base_dirpath = Path("tmp/pages")
     output_base_dirpath.mkdir(exist_ok=True, parents=True)
 
-    dataset_template_fname = collection.attributes.get(
-        "dataset_landing_template", DEFAULT_DATASET_TEMPLATE
-    )
+    dataset_template_fname = DEFAULT_DATASET_TEMPLATE 
+    #dataset_template_fname = collection.attributes.get(
```
liviuba, line 47: Are there are any future uses for the commented code?

kbab: The short answer is I am not sure. Longer answer is: It depends - in the current static site generation, we can use attributes on collections to set things like the template to use. We currently do not have attributes in collections hence it was commented out in the code. It can be uncommented when we have attributes in collections, but we may have moved on to Astro by then. I am happy to remove the comments and put an issue in clickup ...

static-sitegen/scripts/generate_collection_page.py
```diff
@@ -23,18 +24,21 @@
...
     logger.info(f"Loading template {template_fname}")
     logger.info(f"Using dataset page suffix: {page_suffix}")
 
     # Calculate the number of images and number of annotations in a study
     # Put it as an attribute of the study to pass it to collection template
     for bia_study in bia_studies:
-        len_images = len(bia_study.images.keys()) 
+        len_images = len(bia_study.images) 
         bia_study.attributes['no_of_images'] = len_images
         ann_files = get_annotation_files_in_study(bia_study)
```
liviuba, line 43: Does this script run? get_annotation_files_in_study does file_references.values() internally

Can you please give me an example for how to run it? Trying to get rid of load_and_annotate_study :D

liviuba: oh, nevermind, my editor got confused, it's fixed in the PR

static-sitegen/scripts/build_collection_site.py
liviuba: Didn't make sense for them to do the same thing, but now it does. Collections having attributes depends on actually deploying them, I just added a todo to the collections task and will uncomment then

PR #80, opened 2023-12-13
Add changes made to run the pipeline

liviuba:


matthewh-ebi (approved):
LGTM!

tools/snakemake/pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+authors = ["Your Name <you@example.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.10"
+snakemake = "^7.32.4"
+parse = "^1.20.0"
+remotezip = "^0.12.1"
+microfilm = "^0.2.1"
+dev_bia-integrator-core = { path = "../../core/dev", develop = true }
```
ctr26, line 14: great!

core/pyproject.toml
```diff
@@ -14,7 +14,7 @@ ome_types = "^0.3.3"
 requests = "^2.28.0"
 pytest = "*"
 python-jose = {extras = ["cryptography"], version = "^3.3.0"}
-bia-integrator-api = { git = "https://github.com/BioImage-Archive/bia-integrator.git", subdirectory = "clients/python", branch = "biaint-api-backend" }
+bia-integrator-api = { git = "https://github.com/BioImage-Archive/bia-integrator.git", subdirectory = "clients/python" }
```
ctr26, line 17: Does using `{ path = "../clients/python/", develop = true }` like below work?

liviuba: Haven't tested that one specifically, but both worked when I set them up (so it should).

This should be in a readme somewhere, but the logic for these is that what's in the "main" pyproject.toml should be stable / intedend to "just work", and what's under /dev is for when we do a development install where we might change things in multiple packages as we go.

ctr26: Fair, so long as the decision has a reason and is documented

PR #82, opened 2024-01-05
Close annotations

liviuba:
This fixes the issue with annotations already being applied before saving, and adds the apply_annotations request flag & annotations_applied response body flag


ctr26:
Does this address this: https://app.clickup.com/t/86932jp13?

And is this passing API tests?

liviuba:
Yep, all tests pass.

It resolves the one you mentions and this https://app.clickup.com/t/86934exca

api/src/tests/test_file_reference.py
```diff
@@ -42,7 +42,8 @@ def test_create_file_references_multiple_errors(api_client: TestClient, existing
             "uri": "https://test.com/test",
             "size_in_bytes": 1,
             "attributes": {},
-            "annotations": []
```
matthewh-ebi, line 45: Small comment here, but if we always terminate lists with a comma, the diff is a bit easier to read since we're not removing a line, only adding new ones. So in this case, we'd change:

"annotations_applied": False

to:

"annotations_applied": False,

matthewh-ebi (approved):
Looks good to me. Would be great to add something to the example project to show use of the flag, but that can be a separate change.

ctr26 (commented):
Done

PR #83, opened 2024-01-10
Biaint add apply annotations

liviuba:
Adds apply_annotations, json output to biaint. apply_annotations example (needs to be merged first)


matthewh-ebi (approved):
LGTM! It'll be nice to see what's possible with jq for, e.g. summing file sizes. For non-JSON output, I've been enjoying the rich library so we might look at adding that next.

PR #85, opened 2024-02-13
Use custom header with requests library as biostudies rejects standard python header (due to DOS attack)

kbab:
Added a custom header to lines using requests.get to pull data from biostudies as they reject the standard python header due to a DOS attack in December 2023.


kbab:
Changed user-agent text to 'bia-integrator'

AybukeKY (approved):

liviuba (approved):
LGTM!

PR #86, opened 2024-02-16
Fix misc issues to allow snakemake pipeline to run with API

kbab:
bia_integrator_tools/cli.py
Typer does not seem to allow two command line arguments with one being optional, therefore changed accession-id to be an option to allow dual use of image show. If one argument -> image UUID,
or one argument-> image alias and --accession-id option

bia_integrator_tools/omezarrmeta.py
Make 'metadata' field in MultiScaleImage optional as .zattrs from S-BIAD852 did not have this

bia_integrator_tools/rendering.py
Modify assignment of dimension sizes for multiscale images to use info from 'axes' attr as opposed to assuming tczxy order especially when one dimension is missing. Missing dimensions are implicitly assumed to have size=1

scripts/fetch_ome_metadata_for_all_images_in_study.py Add try block to handle case where NGFF has no OME dir

scripts/annotate_study_from_zarr.py
- Created script to add annotation for axes - similar to that for getting dimensions
- Add annotation for axes, and use axes info in metadata when writing sizes of each dimension

### DIFF
--- a/tools/bia_integrator_tools/cli.py
+++ b/tools/bia_integrator_tools/cli.py
@@ -182,13 +182,13 @@ def images_list(accession_id: str, output: OutputFormat = OutputFormat.PRETTY):
 @images_app.command("show")
 def images_show(
         image_uuid_or_alias: str,
-        accession_id: Annotated[Optional[str], typer.Argument(default=None)],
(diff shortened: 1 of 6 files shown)


liviuba (approved):
LGTM!

PR #90, opened 2024-03-05
Model updates

liviuba:
Adds the ImageAcquisition, Specimen, Biosample models.

The main files to look at (non-generated / repetitive) are:
*  `models/persistence.py` where the new models are defined. Now ImageAcquisition doesn't need an actual study/image to exist, which may/may not be good. Same for Specimen and Biosample
* `public.py`/`private.py` for the actual endpoints


clients/python/bia_integrator_api/models/biosample.py
kbab, line 1: The fields organism_scientific_name, organism_common_name, and  organism_ncbi_taxon in the biostudies submission and bia-rembi-models do not have the 'organism_' prefix, but just making the comment - it is not an issue.

The fields experimental_variable, extrinsic_variable and intrinsic_variable are named in the plural in bia-rembi-models and defined as lists of strings. However, in submission they are named in the singular as attributes (hence there is a possibility of more than one for a given biosample?). The uniformity of the names with biostudies makes for easier parsing ...

kbab (approved):
LGTM - left some comments about field names for biosample

PR #93, opened 2024-04-02
updated UUID generation and bulk file and image generation in tests

sherwoodf:
Issue: the tests were more likely to accidentally generate duplicate UUIDs on Macs that linux when batch creating images or files. This was due to the less fine-grained time resolution. Coupled with the api returning 201s when the exact same study is attempted to be created (considered okay, but the behaviour should probably have control flags for it), tests would sometimes not add the expected number of images/files to the DB and fail.

This PR changes how groups of images/files are created in order to 1. make sure they have different UUIDs (by incrementing the input to the generate UUID), and 2. also varies the contents of the metadata, so that a more obvious error will be hit if 2 UUIDs were the same.

Also updates to the readme about setting up VSCode tests, expected poetry errors when setting up the package, and some commands i found useful for testing.

### DIFF
--- a/api/Readme.md
+++ b/api/Readme.md
@@ -1,3 +1,18 @@
+### First time setup
+
+Run
+```
+poetry install
+```
+Will generate the following warning:
+```
+Warning: The current project could not be installed: No file/folder found for package bia-integrator-api
+If you do not want to install the current project use --no-root.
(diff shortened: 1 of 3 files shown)


liviuba (approved):
LGTM!

PR #95, opened 2024-04-05
added json-ld contexts to api response

sherwoodf:
After this change requesting any of the following objects:
- Biosample
- Collection
- ImageAcquisition
- Image
- Specimen
- Study
- StudyFileReference
will return json with an @context field with a URL that points to a file in this github repo. The content type will still be 'application/json', regardless of what gets passed in.

List of changes:

1. Created JSON LD context objects, with some mapping to defined concepts in external ontologies where relevant. Includes ID mappings for objects that should point back to the appropriate BIA api endpoints. Included a generic vocabulary for all undefined terms.

2. Updated models to automatically add in a context object with a link to the github repository path where these files will exist

3. Added test that attempts to parse the json-ld using the contexts, in a similar manner to how i would expect users to parse the json. There is a slight difference, in that they should not need to manually insert the jsonld context (the link should just work), but you can't test changes while pointing to the link. 

4. changed tests that rely on the models to have some placeholder value for the context

5. changed the imports in the tests to avoid `from X import *`

6. Added some type checking & comment to the util method i created for checking unordered lists of dicts were equivalent.

### DIFF
--- a/api/src/models/persistence.py
+++ b/api/src/models/persistence.py
@@ -148,6 +148,8 @@ class BIAStudy(BIABaseModel, DocumentMixin, AnnotatedMixin[StudyAnnotation]):
     file_references_count: int = Field(default=0)
     images_count: int = Field(default=0)
 
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/StudyContext.json")
+
     model_config = ConfigDict(model_version_latest=1)
 
 
@@ -217,6 +219,7 @@ class Biosample(BIABaseModel, DocumentMixin):
     intrinsic_variables: List[str] = Field(
         description="Intrinsic (e.g. genetic) alteration.", default=[]
     )
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/SpecimenContext.json")
 
     model_config = ConfigDict(model_version_latest=1)
 
@@ -229,6 +232,7 @@ class Specimen(BIABaseModel, DocumentMixin):
     )  # is this a ST-only concern, or does it make sense for it to be in the models?
     sample_preparation_protocol: str = Field()
     growth_protocol: str = Field()
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/SpecimenContext.json")
 
     model_config = ConfigDict(model_version_latest=1)
 
@@ -248,6 +252,7 @@ class ImageAcquisition(BIABaseModel, DocumentMixin):
     imaging_method: str = (
         Field()
     )  # make this an Enum / restrict some other way? Distinguishing between "somewhat close to a controlled vocabulary" vs "completely free text" might be useful
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/ImageAcquisitionContext.json")
 
     model_config = ConfigDict(model_version_latest=1)
 
@@ -278,6 +283,7 @@ class BIAImage(BIABaseModel, DocumentMixin, AnnotatedMixin[ImageAnnotation]):
         description="Context in which the image was acquired. This list often has one item, but it can occasionally have more (e.g. for multimodal imaging)",
         default=[],
     )
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/ImageContext.json")
 
     model_config = ConfigDict(model_version_latest=2)
 
@@ -291,6 +297,7 @@ class BIACollection(BIABaseModel, DocumentMixin, AnnotatedMixin[CollectionAnnota
     subtitle: str = Field()
     description: Optional[str] = Field(default=None)
     study_uuids: List[str] = Field(default=[])
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/CollectionContext.json")
 
     model_config = ConfigDict(model_version_latest=1)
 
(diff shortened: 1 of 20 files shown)


api/src/models/persistence.py
```diff
@@ -278,6 +283,7 @@ class BIAImage(BIABaseModel, DocumentMixin, AnnotatedMixin[ImageAnnotation]):
         description="Context in which the image was acquired. This list often has one item, but it can occasionally have more (e.g. for multimodal imaging)",
         default=[],
     )
+    context: str = Field(alias='@context', default="https://github.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/ImageContext.json")
```
liviuba, line 286: I think these something like https://raw.githubusercontent.com/BioImage-Archive/bia-integrator/tree/main/api/src/models/jsonld/1.0/ImageAcquisitionContext.json (otherwise you get html instead of the raw file)

api/src/tests/test_rdf.py
```diff
@@ -0,0 +1,51 @@
...
+    [
+        ( 'CollectionContext.jsonld', get_template_collection(add_uuid=True) ),
+        ( 'StudyContext.jsonld', get_template_study(add_uuid=True) ),
+        ( 'ImageContext.jsonld', get_template_image({"uuid": uuid.uuid4()}, add_uuid=True) ),
+        ( 'StudyFileReferenceContext.jsonld', get_template_file_reference({"uuid": uuid.uuid4()}, add_uuid=True) ),
+        ( 'BiosampleContext.jsonld', get_template_biosample(add_uuid=True) ),
+        ( 'SpecimenContext.jsonld', get_template_specimen({"uuid": uuid.uuid4()}, add_uuid=True) ),
+        ( 'ImageAcquisitionContext.jsonld', get_template_image_acquisition({"uuid": uuid.uuid4()}, add_uuid=True) ),
+    ]
+)
```
liviuba, line 34: This tripped me up (thinking that a uuid is added in arg1 and then the flag set anyway in arg2)

Generally, get_template_* return dicts with template things, and existing_* return existing object (created in the api). In this case just using the existing_file_reference fixture would've made sense, but turns out pytest has issues with parameterised fixtures (https://github.com/pytest-dev/pytest/issues/349). There is https://github.com/tvorog/pytest-lazy-fixture but not sure if worth it

Maybe consider adding a comment at the top of this block to explain this? It would at least make it easy(er) to find the example if we end up adding pytest-lazy-fixture sometime later. Or just add pytest-lazy-fixture as a dependency and we just have it/use it for new tests?

ctr26:
This is an unfinished thought but I think I should note here anyway. 

I'd prefer if the contexts string field(s) were constructed by a class. Either through composition or mixin, though I think my preference so far would be a composition context class that has a __repr__ that returns a string. A mixin would need a post_init function which might clash with other future inheritors.

My reason being repeating the same string multiple times is error prone in future development where a developer might change one field correctly but miss another not another. Moreover there might be cases where you want the url to change (maybe to a file:/// for local development or something) and then that should be changed in the .env file ( eventually not today). It also has the advantage of future version bumps being smaller PRs.

Also, using the raw. GitHub path is better than the tree one. For pure file serving a common pattern I've seen is using the GitHub.io path. This is what is done for helm file serving, but in this case I don't see any meaningful advantage.

Thoughts? If be interested in knowing if my above points are moot because of the nature of the context.

sherwoodf:
> I'd prefer if the contexts string field(s) were constructed by a class. Either through composition or mixin, though I think my preference so far would be a composition context class that has a **repr** that returns a string. A mixin would need a post_init function which might clash with other future inheritors.
> 
> My reason being repeating the same string multiple times is error prone in future development where a developer might change one field correctly but miss another not another. Moreover there might be cases where you want the url to change (maybe to a file:/// for local development or something) and then that should be changed in the .env file ( eventually not today). It also has the advantage of future version bumps being smaller PRs.
> 
> Thoughts? If be interested in knowing if my above points are moot because of the nature of the context.

Definitely not moot. I had initially create a base dictionary which had some of the content used in all contexts (vocab, title) etc which was then used in the more specific ones which i used to create the json directly in objects. When i switched to the json file i removed that 'hierarchy'.

While I agree that the current system would be prone to errors with changes, i don't think making a better one is worth development time. It feels too early to try to create the 'correct' system as i'm not confident that we know what result we're aiming for. All the values should currently be conisdered placeholders: most of the important ones don't have an existing definitions, and where i have created a mapping, i'd bet against changes in one file needing to be reflected in all the others. I'd feel more comfortable coming back to it in ~4 months time to create a system that reflects the data-modelling decisions of the working group.

I view this more change as getting the api to return valid JSON-LD to make valid RDF to play with, but no where near the 'correct' RDF that conforms to an ontology of our api concepts (which still needs defining).

ctr26:
Yep, valid reasons.

If the structure is being actively developed and is prone to significant change anyway then paying some technical debt now is your decision.

Thanks for the clear and thoughtful response

api/src/models/persistence.py
```diff
@@ -62,6 +68,15 @@ def __init__(self, *args, **data):
         super().__init__(*args, **data)
 
 
+class JSONLDDocument(DocumentMixin, BaseModel):
```
liviuba, line 71: This makes sense, I'm just wondering if we would ever want jsonld schemas for nested objects (e.g. for ImageRepresentation) in the relatively-near future.

This encourages assuming that JSONLDDocuments are DocumentMixins in future code, which might be tricky to separate if we ever want @schemas for non-base objects

api/src/tests/util.py
```diff
@@ -455,5 +455,7 @@ def assert_bulk_response_items_correct(
                 assert rsp.status_code == 404
 
 
-def tests_base() -> str:
```
liviuba: I get a test discovery error for this because it tries to import tests_base from test_image.py (for the ome xml file)

liviuba (changes requested):
Looks good, please check suggestions

api/src/tests/util.py
sherwoodf: fixed in new commit

api/src/models/persistence.py
sherwoodf: switched to jsonld mixon

liviuba (approved):
LGTM!

PR #96, opened 2024-04-09
Fix 500 response for invalid ome xml

liviuba:
This came up as a bug yesterday. Invalid content is a client-side error so should have a 4xx status


sherwoodf (commented):
Approved

PR #97, opened 2024-04-10
Add readonly client builder

liviuba:
Add readonly client builder, needed for nice explorer code (wip).

Some things ignored, to decide next wek
* do we keep simple_client (the private version) named as it is? get_client and simple_client in the same file might be confusing, but this means a rename everywhere we use a client
* do we add the api server as-is in the LB so we can get rid of the hardcoded ip and no host checking (get_client assumes yes, for tidyness)

### DIFF
--- a/clients/python/bia_integrator_api/util.py
+++ b/clients/python/bia_integrator_api/util.py
@@ -8,6 +8,8 @@ import json
 import hashlib
 import uuid
 
+PUBLIC_API_URI = "https://45.88.81.209:8080"
+
 def get_access_token(api_config: Configuration, username: str, password: str) -> str:
     api_client = ApiClient(configuration=api_config)
     default_api = PrivateApi(api_client=api_client)
@@ -67,7 +69,7 @@ def simple_client(
     return client
 
 def get_client(
-    api_base_url: str = "https://45.88.81.209:8080"
+    api_base_url: str = PUBLIC_API_URI
 ) -> PublicApi:
     api_config = Configuration(
         host = api_base_url
--- a/clients/python/bia_integrator_api/util.py
+++ b/clients/python/bia_integrator_api/util.py
@@ -1,5 +1,5 @@
 from bia_integrator_api import Configuration, ApiClient
-from bia_integrator_api.api import PrivateApi
+from bia_integrator_api.api import PrivateApi, PublicApi
 from base64 import b64decode
 from typing import Optional
 import time
@@ -65,6 +65,18 @@ def simple_client(
     client = PrivateApi(api_client=api_client)
 
     return client
+
+def get_client(
+    api_base_url: str = "https://45.88.81.209:8080"
+) -> PublicApi:
+    api_config = Configuration(
+        host = api_base_url
+    )
+    api_config.verify_ssl = False
+    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+    api_client = ApiClient(configuration=api_config)
+    return PublicApi(api_client=api_client)
 
 def uuid_from_str(doc_stable_string: str) -> str:
     hexdigest = hashlib.md5(doc_stable_string.encode("utf-8")).hexdigest()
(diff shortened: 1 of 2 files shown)


clients/python/bia_integrator_api/util.py
```diff
@@ -66,6 +66,18 @@ def simple_client(
 
     return client
 
+def get_client(
```
sherwoodf, line 69: Maybe call this 'read_only_client', instead of 'get_client'. I parse this as a function that returns a client (i.e. it 'gets' me a client). But let me know if this is standard nomenclature for these kinds of objects.

sherwoodf (changes requested):

clients/python/bia_integrator_api/util.py
liviuba: It does return a client, the reason I wanted to avoid something like read_only_client is that it implies there is some non-readonly client, which isn't relevant for someone that just wants to fetch some data, and may cause confusion (from the perspective of a data-focused user, that just wants something to pull data with). This should be the most common usecase for someone that isn't internal to the team.

This is why I would have rather called this something as generic as possible (but maybe not get_client) and rename the other one to something like get_rw_client or similar, so the extra detail of having two clients, one for ro and one for rw is only shown in a context where it's meaningful. This may be extra picky / just add inconsistency though, @ctr26 do you have a preference?

sherwoodf: Ah so get client is 'get me a client' and just happens to also be 'a client that can only make get requests'?

clients/python/bia_integrator_api/util.py
```diff
@@ -66,6 +66,18 @@ def simple_client(
...
+    api_base_url: str = "https://45.88.81.209:8080"
```
ctr26, line 70: I'd maybe declare this urI at the top as 

PUBLIC_API_URI

or something and then 

api_base_url: str = PUBLIC_API_URI

ctr26: The pattern I prefer instead of factory functions like this would be a simpler class that inherits, with defaults and logic handled at instantiation:

@dataclass
class PublicClient(PublicApi):
    api_base_url: str = "https://45.88.81.209:8080"
    verify_ssl: bool = False
    api_client: ApiClient = field(init=False)

    def __post_init__(self):
        config = Configuration(host=self.api_base_url, verify_ssl=self.verify_ssl)
        if not self.verify_ssl:
            urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
        self.api_client = ApiClient(configuration=config)
        super().__init__(api_client=self.api_client)

ctr26: N.B I've not checked if this worked I just free typed it

clients/python/example/readonly_start_here.py
```diff
@@ -0,0 +1,8 @@
+from bia_integrator_api.util import get_client
```
ctr26, line 1: See comment on using class inheritance to build simple objects above. Avoids needing a utils module for dumping random functions

clients/python/bia_integrator_api/util.py
liviuba: I agree the class-based way is nicer, but I was trying to avoid inheriting the client to force use of the only public api class we generate and discourage adding extra fields on the child class which would make them (and the models) diverge and create confusion. We might end up doing this anyway if we pull some of the "wrapper" part of bia-explorer into the client, but I'm not sure we would/should.

Also, adding a class wrapper for the RO client but not RW isn't nice, but to add it to the RW one we would need to update the client code to use it (this was also the reason for the odd naming, simple_client and get_client), or keep both for a while and deprecate simple_client.

I would rather add this as a new ticket, to decide on how much of the wrapper in bia-explorer we move back to the api client, and also how to deal with the breaking change (or deprecation of simple_client)

ctr26 (approved):

PR #98, opened 2024-04-10
added Api paramenter to control idempotent write

sherwoodf:
Added an api parameter that enables control of allowing idempotent writes. Flag is by default off, so the API will return 400 if two identical entries are attempted to be written at the same time, but can be set to True to enable this to pass (which might be useful for bulk updates to documents).

Ticket: https://app.clickup.com/t/86947j7ap

Updated existing tests to account for the flag, testing some cases explicitly with the flag off and then on. We could add more tests, but would like to go through a separate test refactor to ensure more general/consistent coverage across api paths. I do update some some tests to try to name them consistently based off the api endpoint name, as most do this.


kbab (approved):
LGTM

api/src/api/private.py
```diff
@@ -75,6 +80,7 @@ async def create_images(
         db.find_study_by_uuid,
         ref_bulk_operation_response=create_bulk_response,
     )
+    db.idempotent_create = idempotent_create
```
liviuba, line 83: If this is set, could it always be at the very top of the request handler? Just to make it more obvious that no db calls ever happen without this flag being set

api/src/api/private.py
```diff
@@ -212,9 +226,12 @@ async def update_image_acquisition(
 
 @router.post("/specimens", status_code=status.HTTP_201_CREATED)
 async def create_specimen(
-    image_acquisition: db_models.Specimen, db: Repository = Depends()
+    image_acquisition: db_models.Specimen,
+    db: Repository = Depends(),
+    idempotent_create=False,
 ) -> None:
     logging.info(f"Creating specimen {image_acquisition.uuid}")
+    db.idempotent_create = False
```
liviuba, line 234: typo for db.idempotent_create = idempotent_create?

liviuba (commented):
Looks good, please see comments

api/src/models/repository.py
```diff
@@ -19,11 +20,26 @@
 COLLECTION_OME_METADATA = "ome_metadata"
 
 
+class OverwriteMode(str, Enum):
+    FAIL = "fail"
+    ALLOW_IDEMPOTENT = "allow_idempotent"
+
+    # this enables FAIL to be mapped to fail.
+    @classmethod
```
liviuba, line 28: I'm not sure if supporting this kind of behaviour (synonyms for API params) is more confusing than helpful, especially since we don't support it for other things.

IMO clear rules for what the attribute should look like exactly (either capitalised or not) are simpler to follow than things like "some fields we acept capitalised or not, other fields we acept either way but actually cast to PascalCase", etc. And the actual values would be constants in the generated client anyway

api/src/tests/test_misc.py
```diff
@@ -136,3 +126,70 @@ def test_single_doc_update_some_change_fails(api_client: TestClient, existing_st
...
+        ),
+    ],
+)
+def test_idempotent_create(
+    api_client: TestClient,
+    endpoint: str,
+    document_creator: str,
+    request,
+):
+    document = request.getfixturevalue(document_creator)
```
liviuba, line 155: It's just one line, but I think between this looking messier (just because we have to do 2 things to get a fixture in a test) and losing the ability to ctrl-click into fixtures since they're now strings, it's nicer to just add the lazy fixtures dev dependency

Plus, this would get used a lot if we group tests by functionality and deduplicate (as we go along), and doing this wouldn't jump to my mind 1 month from now

api/src/tests/test_image.py
```diff
@@ -434,7 +444,9 @@ def test_post_invalid_ome_metadata(api_client: TestClient, existing_image: dict)
     assert rsp.status_code == 404
 
 
-def test_image_ome_metadata_update(api_client: TestClient, existing_image: dict):
+def test_set_image_ome_metadata_updateSylvaneth27(
```
liviuba, line 447: Warhammer typo?

api/src/tests/conftest.py
```diff
@@ -0,0 +1,87 @@
+from fastapi.testclient import TestClient
```
liviuba, line 1: This is cool, but maybe consider adding a comment along the lines of "Automatically imported everywhere by pytest" for people like me that didn't know this is a thing / are twichy about framework magic?

liviuba (commented):
Looks good, please see comments

api/src/models/repository.py
sherwoodf: Any preference for capitalised or not?

ctr26: In python 3.11 you can do string enums. That would be my preference but I'd rather not enforce 3.11

liviuba (approved):
LGTM!

PR #99, opened 2024-04-16
New model fixes

liviuba:
This is for https://app.clickup.com/t/869437z42 . It's a small tweak to remove `_methods_` from `image_acquistions_uuids` (and the updated examples.

Needed to use the latest client in explorer code


sherwoodf (approved):
LGTM

PR #101, opened 2024-04-19
Fix issue causing CI/CD failures

kbab:
Since the API started being used we had some CI/CD tests that were failing because they were based on the old invocation using data in the data_dirpath.

This PR fixes this by updating tests to use the RO version of the API and deleting a couple of tests that do not work smoothly with the API.


ctr26:
Good changes towards using proper pytest fixtures and generalising the testing overall.

Net negative PR 👍

Any way around using the environmental variables using dotenv or something? Tbh I can't immediately see how the env vars are used

liviuba (approved):
LGTM!

PR #103, opened 2024-04-22
Fix race condition when db gets reused

liviuba:
This is deployed already, there was an issue with a 500 error when doing 5(parallel)*44 (sequential) requests getting all images of S-BIAD582. I think somewhere in there a cached db client was reused across requests (maybe for a pending request), but we explicitly close the connection before applying (or not) any annotations.

Updated from the initial PR to only close the connection if apply_annotations=True (so in that case, the 500 error would still happen, but we don't regularly use that now)

### DIFF
--- a/api/src/models/repository.py
+++ b/api/src/models/repository.py
@@ -87,6 +87,13 @@ class Repository:
         self.close()
 
     def close(self):
+        """
+        When we apply_annotations, we double-close.
+        Ignore the double-close exception
+
+        See https://motor.readthedocs.io/en/stable/api-asyncio/asyncio_motor_client.html#motor.motor_asyncio.AsyncIOMotorClient.close
+        """
+
         try:
             self.connection.close()
         except InvalidOperation:
--- a/api/src/models/repository.py
+++ b/api/src/models/repository.py
@@ -13,6 +13,7 @@ from enum import Enum
 import uuid
 import pymongo
 import os
+from pymongo.errors import InvalidOperation
 
 DB_NAME = os.environ["DB_NAME"]
 COLLECTION_BIA_INTEGRATOR = "bia_integrator"
@@ -86,7 +87,10 @@ class Repository:
         self.close()
 
     def close(self):
-        self.connection.close()
+        try:
+            self.connection.close()
+        except InvalidOperation:
+            pass
 
     async def file_references_for_study(
         self, *args, **kwargs
--- a/api/src/models/repository.py
+++ b/api/src/models/repository.py
@@ -89,15 +89,12 @@ class Repository:
     def close(self):
         """
         When we apply_annotations, we double-close.
-        Ignore the double-close exception
+        Ignore the double-close exception?
 
         See https://motor.readthedocs.io/en/stable/api-asyncio/asyncio_motor_client.html#motor.motor_asyncio.AsyncIOMotorClient.close
         """
 
-        try:
-            self.connection.close()
-        except InvalidOperation:
-            pass
+        self.connection.close()
 
     async def file_references_for_study(
         self, *args, **kwargs
--- a/api/src/api/annotator.py
+++ b/api/src/api/annotator.py
@@ -37,11 +37,10 @@ class Annotator:
             db_models.AnnotatedMixin | List[db_models.AnnotatedMixin]
         ) = False,
     ):
-        self.db.close()
-
         if not self.apply_annotations:
             return
 
+        self.db.close()
         if isinstance(response_object, list):
             for rsp_item in response_object:
                 self._apply_annotations(rsp_item)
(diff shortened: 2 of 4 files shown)


liviuba:
> And my understanding is that `Depends(use_cache=True),` is just there to improve performance rather than fix the race condition yes?

I'd have to double check, but IIRC use_cache=True (the default) causes the bug, because that's why Starlette tries to reuse the same `db` dependency for two requests and the race condition happens. I think the actual fix is just this line (https://github.com/BioImage-Archive/bia-integrator/pull/103/files#diff-d256968daf44a4ab334135b7693ef710eb8f1b34b26adc80b008874885999a96L39) 

Rebasing / re-checking now

liviuba:
I think I just changed what actually fixes it (the double close), but I really need it merged & deployed to be sure

api/src/models/repository.py
```diff
@@ -86,7 +87,17 @@ def __exit__(self, *_):
         self.close()
 
     def close(self):
-        self.connection.close()
+        """
+        When we apply_annotations, we double-close.
```
sherwoodf, line 91: I'm a little bit confused about the 'why' of this PR, so i'm just going to write my understanding (/ guesses) & please correct me. This isn’t about the code in your PR, it's more just me not understanding the context for the change & therefore why this double close is being called out.

Assumption: we want to close the db connection at least once at some point. By double-close you mean calling close twice & getting an InvalidOperation the 2nd time.

Current state:
There is a situation in which we might return a 500 because a db client that was being reused across multiple requests was having it’s connection closed prior to all the requests it was trying to handle being completed. We are closing the connection twice all the time, once explicitly, once with some ‘magic’ logic around ‘async repository_dependency()’ (by magic i mean some combination of async & logic that its the finally in that function). We currently close the connection a 2nd time/explicitly when calling 'annotated_if_needed' regardless of whether or not we actually want to annotate.

Expected state after this PR:
This changes it the logic to only explicitly close the connection after requests with apply_annotations=True. We still rely on the ‘magic’ logic of repository_dependency to close the connection normally, but with apply_annotations=True we call close twice. We might still get the 500, but since we don't use apply_annotations very often we don't expect to making bulk requests with it to get this accidental closing of unfinished requests behaviour.

Basically i can only see 1 call of close() in the api code & you’ve changed it to only happen after the ‘if’ logic, so i’m assuming it must get closed elsewhere for you to want to explicitly call out that apply_annotations=True means we try to close it twice?

liviuba: This is both a bit old and slightly handwavy because I didn't dig too deep into the fastapi code, but what basically triggered this were tests in one of the tools being flakey, because they are pointed at the prod api. What would happen is they would fail initially, but then if you rerun them they work.

Traced it back to a case where if you have 1 request being processed and one in the pending queue, FastAPI creates the db dependency once, and then reuses the same dependency (without recreating it) when it pops the queued request.

We close the db connection in either
* `Annotatos::annotate_if_needed` (which we don't use now anyway, but will in the future), to make sure that if we ever do something like a fetch-update-push loop in the api there is no chance we write back with annotations applied.
* or `app.py::repository_dependency` in the finally block. The function yields the db handle when the dependency gets built, but doesn't get control again for every request, so it can't run the finally block (I think)

So the solution is to only close the connection when we apply annotations (so we don't close early a resource that we don't have control over and fastapi might decide to reuse when handling the next request), and also ignore the double-close error if it happens.

This would still not behave well if we had 1 request with apply_annotations=true (so we'd close the db) and then another one queued with false (so it would use the db, which would be reused, and so it couldn't). We would probably get an exception somewhere in that case, though, but I could probably be more precise about the skipped exception

liviuba: Tried to get it to raise an exception and couldn't, I think Motor (or Mongo) doesn't actually double-close, so if you close something that's closed already it just ignores it (at least when I tested it).

Docs say otherwise, but I'd rather not ignore an exception and see if we still get the issue, so I deleted the try-catch but left the docstring in case we need it

sherwoodf: ;Okay cool! Sounds like i had understood what was going on. One extra question though, you said:
> Annotatos::annotate_if_needed (which we don't use now anyway, but will in the future)

But we do use it now right? It gets called by all the endpoint code.

sherwoodf (approved):

api/src/models/repository.py
liviuba: I mean we don't get any requests with apply_annotations=True, so we call annotate_if_needed but we don't actually apply any annotations (now)

sherwoodf: Ah right. I think i used it the other week! But that was mostly to see what it would do...  So yes, i get what you mean - i wouldn't expect it to be used a lot.

liviuba: The actually nice solution to this would be to set use_cache=False for that dependency in particular, but then we don't reuse connections when we can which isn't nice. There are probably better ways to invalidate the object but this is cheap

PR #104, opened 2024-04-25
Add initial version of bia_ingest package

kbab:
Add bia_ingest package to bia-integrator. This is current state of ingest code that is WIP. This commit has functionality to attach biosample, specimen and image acquisition to images.


bia-ingest/bia_ingest/conversion.py
```diff
@@ -0,0 +1,409 @@
...
+    # Then we can flatten this list of lists:
+    flattened = sum(nested, [])
+
+    for section in flattened:
+        find_sections_recursive(section, search_types, results)
+
+    return results
+
+
+def find_sections_in_submission(submission: Submission, search_types: List[str]):
```
sherwoodf, line 55: Is this method necessary? can't you just call find_scections_recursive?

bia-ingest/bia_ingest/conversion.py
```diff
@@ -0,0 +1,409 @@
...
+    """Find all sections within the submission with the given type."""
+
+    return find_sections_recursive(submission.section, search_types, [])
+
+
+def mattributes_to_dict(attributes: List[Attribute], reference_dict: Dict) -> dict:
```
sherwoodf, line 61: why is this called 'mattributes'? (is it because it has some mapping)?

bia-ingest/bia_ingest/conversion.py
```diff
@@ -0,0 +1,409 @@
...
+        reference_dict[section.accno] = Affiliation(**model_dict)
+
+    key_mapping = [
+        ("name", "Name", None),
+        ("email", "E-mail", None),
+        ("role", "role", None),
+        ("affiliation", "affiliation", None),
+        ("orcid", "ORCID", None),
+    ]
+    author_sections = find_sections_in_submission(submission, ["author"])
```
sherwoodf, line 212: Could this use find_authors_in_submission?

bia-ingest/bia_ingest/scli.py
```diff
@@ -0,0 +1,36 @@
+import logging
```
sherwoodf, line 1: ... do we now know what scli stands for? XD

bia-ingest/tests/src/test_visitor_finds_value_qualifiers.py
```diff
@@ -0,0 +1,42 @@
+from bia_ingest.visitor import Visitor
+
+submission = {
```
sherwoodf, line 3: I like defining the example in the same file as the tests that use it as it makes it easier to read

bia-ingest/bia_ingest/scli.py
sherwoodf: Also, where is the cli file?

bia-ingest/bia_ingest/conversion.py
kbab: The method seems to simply call find_sections_recursive - so it seems redundant and I am happy to remove it. However, I will check with @matthewh-ebi as he wrote it. The only reason I can think of is to have a stable interface for finding sections in submissions that allows changing the implementation without having to modify the callers.

bia-ingest/bia_ingest/conversion.py
kbab: Thanks for picking up on this. `find_and_convert_authors` means `author_section_to_author` and `find_authors_in_submission` are redundant. I will refactor

sherwoodf (approved):

bia-ingest/pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[tool.poetry.dependencies]
+python = "^3.11"
+requests = "^2.31.0"
+typer = "^0.9.0"
+rich = "^13.7.0"
+fs = "^2.4.16"
+python-dotenv = "^1.0.0"
+bia-rembi-models = {git = "https://github.com/BioImage-Archive/bia-rembi-models.git"}
+bia-integrator-core = {path = "../core"}
```
liviuba, line 17: I thought we were deprecating core (at least in the sense of not adding new code that uses it if not migrating existing code). When it was moved over to the api, the main constraint was to not change anything downstream, not to actually make it work properly.

Not sure if this is has high enough scope that it's worth a whole design meeting, but we shoul definitely discuss how to move some of the explorer wrapper into the client as helpers and get rid of core altogether, than add any new dependencies to core IMO.

bia-ingest/bia_ingest/visitor.py
```diff
@@ -0,0 +1,175 @@
...
+
+import re
+import uuid
+import hashlib
+from .models import BIAStudy, Publication, Author, Affiliation
+from bia_integrator_api.models.biosample import Biosample
+from .conversion import generate_biosample_uuid
+
+
+class Visitor:
```
liviuba, line 14: This (and the book) are cool, but I'm really confused about what it does.

My understanding is that it takes in json-pagetab, does a jsonquery-like thing (with visit), and gives back api models-shaped objects? The Biosample part is really useful because fields that actually need to be parsed to get api-shaped things didn't cross my mind.

I'm not sure this is the simplest solution though. I thought there was some code to parse (json-pagetab) dicts into (json-pagetab-shaped) objects, can't we add a `to_api_model` to those classes to handle the model translation from pydantic-json-pagetab to actual-api-model (e.g. things like the split for common/scientific name)? This would at least let pydantic deal with json-pagetab to object-json-pagetab conversion, and also isolate the "mapping" to individual classes, making parsing code more readable. Not 100% sure what I think exists for pagetab parsing actually exists though.

bia-ingest/tests/src/test_assign_image_acquisitions_to_images_script.py
```diff
@@ -11,43 +11,48 @@
...
+        ["accession_id", "path", "size"],
     )
     test_filerefs.append(fileref)
 
+
 def mock_get_study_images():
     images = []
     for i in range(N_TEST_IMAGES):
-        image_rep = type('', (), {})()
+        image_rep = type("", (), {})()
```
ctr26, line 52: Is this empty object  doing to the same job as Simplenamespace?

https://docs.python.org/3/library/types.html#types.SimpleNamespace

kbab: Thanks @ctr26 . Yes it is - I was not aware of SimpleNamespace. However, I have changed to using 'Mock' as I think it fits this usecase better ...

liviuba:
My understanding was that the guideline for updating with main was with rebasing instead of merging (unless an exception), this has some of the merged changes from a different branch, which makes the PR harder to understand. Is it intentional?

kbab:
> My understanding was that the guideline for updating with main was with rebasing instead of merging (unless an exception), this has some of the merged changes from a different branch, which makes the PR harder to understand. Is it intentional?

No - it was not intentional. I got the workflow wrong - apologies.

bia-ingest/tests/src/__init__.py
sherwoodf, line 1: Why is there a src/ folder inside tests? Would have expected it the other way around.

ctr26: Bothering me too, I was looking for out style guidelines, my preferred test structure is where you match the modules in the test dir

i.e.
```
module_a.py
module_b.py
tests/
  ├── conf_test
  ├── module_a.py
  ├── module_b.py

```

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,244 @@
...
+            if image_acquisitions_uuid != image.image_acquisitions_uuid:
+                image.image_acquisitions_uuid = image_acquisitions_uuid
+                image.version += 1
+                api_client.update_image(image)
+                logger.info(
+                    f"Setting image {image.uuid} image_acquisitions_uuid to : {image_acquisitions_uuid}"
+                )
+
+
+@click.command()
```
sherwoodf, line 219: was there something that click does that typer doesn't that this code needs? I don't have anything against click, it's just we use typer for the bia agent and i'd rather not have to learn how to use one cli library.

ctr26: typer wraps click I believe

bia-ingest/tests/src/__init__.py
kbab: I included the `test/src` directory because I had some test data in `test/resources` and wanted that level to only have directories, not test modules and the `resources` directory . However I cannot find any online resources that backup this structure. I will move the test modules up one level and also change the name of the `resources` directory to `data` to align with the name in the `api` package

bia-ingest/scripts/assign_image_acquisitions_to_images.py
kbab: Thanks for raising this. The scripts in `tools/scripts` and `static-sitegen/scripts` (which will be deprecated when we start using Astro) use `click` . So as a group we need to decide which to use. I am happy with either. I will mention this in code-update-and-review slack channel and use whatever we agree on.

bia-ingest/bia_ingest/visitor.py
kbab: No it wasn't. I did not introduce using the `Visitor` object, but left this piece of code in-place as I use it for other purposes - e.g. traversing biostudies submissions in the script to compare artefacts in API studies with those in their equivalent Biostudies submissions. However, this is on a different branch.

bia-ingest/scripts/assign_image_acquisitions_to_images.py
kbab: There are three responses in code-review-and-update channel supporting `typer`. Will change to use `typer`

bia-ingest/tests/test_conversion_author_functions.py
```diff
@@ -0,0 +1,36 @@
...
+        email="test_author1@ebi.ac.uk",
+        role="corresponding author",
+        affiliation=Affiliation(name="Test College 1"),
+        orcid="0000-0000-0000-0000",
+    )
+    return author
+
+
+@pytest.fixture
+def expected_author2():
```
sherwoodf, line 21: This is simple enough, and wants different values for the fields but this solution: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#factories-as-fixtures 
seems to be the official 'want to call the same fixture multiple times' solution

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,246 @@
...
+    def _find_element_by_title(title):
+        def _inner(item):
+            return item.get("title", "") == title
+
+        return _inner
+
+    image_acquisition_uuids = []
+    for association in associations:
+        # ToDo: Check if there can be > 1 entry with same title? YES!!!
+        # ToDo: Refactor code blocks below to iterate using a dict
```
sherwoodf, line 70: (super nit pick) but in general i think TODO get's picked up by syntax highlighting, but ToDo doesn't.

bia-ingest/bia_ingest/biostudies.py
```diff
@@ -0,0 +1,325 @@
...
+    template which can be populated with the value of relpath to produce
+    the URI for a given file."""
+
+    request_uri = f"https://www.ebi.ac.uk/biostudies/api/v1/studies/{accession_id}/info"
+    r = requests.get(request_uri)
+    raw_obj = json.loads(r.content)
+    # Strip the initial ftp from the ftp link, replace by http and add /Files
+    accession_base_uri = "https" + raw_obj["ftpLink"][3:] + "/Files"
+
+    file_uri_template = accession_base_uri + "/{relpath}"
```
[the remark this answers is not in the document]
kbab: It is a left over from the original code that was copied over - and will probably be used once bia-ingest is fully implemented. It allows formating of a path i.e `file_path = file_uri_template.format(relpath="path_to_file.png")`.

bia-ingest/pyproject.toml
kbab: Yes - we agreed to deprecate core. All references to bia-integrator-core have been removed from bia-ingest package.

bia-ingest/bia_ingest/models.py
```diff
@@ -0,0 +1,36 @@
+from typing import List
+
+from pydantic import BaseModel
```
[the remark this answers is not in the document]
kbab: Yes - the API models need to be changed, which will happen during this OKR cycle and the changes will be reflected here.

bia-ingest/pyproject.toml
```diff
@@ -0,0 +1,25 @@
...
+packages = [{include = "bia_ingest"}]
+
+[tool.poetry.dependencies]
+python = "^3.11"
+requests = "^2.31.0"
+typer = "^0.9.0"
+rich = "^13.7.0"
+fs = "^2.4.16"
+python-dotenv = "^1.0.0"
+bia-integrator-api = { git = "https://github.com/BioImage-Archive/bia-integrator.git", subdirectory = "clients/python" }
```
liviuba, line 16: 👍

bia-ingest/tests/test_assign_image_acquisitions_to_images_script.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def test_assign_image_acquisitions_to_images_script_runs_ok(caplog):
+    caplog.set_level(
+        logging.INFO, logger=assign_image_acquisitions_to_images.logger.name
+    )
+    runner = CliRunner(mix_stderr=False)
+    result = runner.invoke(app, [accession_id,], catch_exceptions=False)
+
+    if result.exception is not None:
```
liviuba, line 108: Not tested but looking at the docstring for invoke, if catch_exceptions=False and there is an exception, this never gets reached, same for the `result.exception` assert below

This is good IMO because exceptions aren't expected anyway so they can just propagate cleanly to the top to help with debugging, but then this and the assert aren't used and could be deleted?

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,246 @@
...
+global api_client
+api_client = simple_client(
+    os.environ.get("BIA_API_BASEPATH", "http://localhost:8080"),
+    os.environ.get("BIA_USERNAME", ""),
+    os.environ.get("BIA_PASSWORD", ""),
+    os.environ.get("BIA_DISABLE_SSL_HOST_CHECK", ""),
+)
+
+
+def image_acquisitions_from_associations(submission, associations):
```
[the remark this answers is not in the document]
kbab: Added type hints

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,246 @@
...
+    """Create/retrieve image acquisition models for specified associations"""
```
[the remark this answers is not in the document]
kbab: Yes, it returns a list of uuids (but they are of type string). It checks biosample and specimen because if ImageAcquisition UUID does not exist already in API it has to be created and ImageAcquisition encapsulates Specimen which encapsulates Biosample .

bia-ingest/scripts/assign_image_acquisitions_to_images.py
kbab: Thanks - will change.

bia-ingest/bia_ingest/biostudies.py
```diff
@@ -0,0 +1,325 @@
...
+
+class Section(BaseModel):
+    type: str
+    accno: Optional[str]
+    attributes: List[Attribute] = []
+    subsections: List[Union["Section", List["Section"]]] = []
+    links: List[Link] = []
+    files: List[Union[File, List[File]]] = []
+
+    def as_tsv(self, parent_accno=None):
```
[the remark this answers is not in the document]
kbab: Yes! I suspect it was copied over from there. This and other redundancies should be reduced or eliminated once we agree on the models to use for the website MVP. I am happy to discuss how we keep track of the issues already identified.

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,255 @@
...
+        "zipped_zarr",
+    ]
+    for image in images:
+        if image.image_acquisitions_uuid is None:
+            image.image_acquisitions_uuid = []
+        image_acquisitions_uuid = []
+        image_reps = [
+            i
+            for i in filter(
+                lambda i: i.type in bia_image_types and "attributes" in i.__dict__,
```
liviuba, line 206: Same for this block, can it just be (more readable IMO, fewer possible paths & clearer condition for update)

```
if set(image.image_acquisitions_uuid) != set(image_acquisitions_uuid):
    image.image_acquisitions_uuid = image_acquisitions_uuid
    image.version += 1
    api_client.update_image(image)
    logger.info(
        f"Setting image {image.uuid} image_acquisitions_uuid to : {image_acquisitions_uuid}"
    )
```

bia-ingest/tests/test_assign_image_acquisitions_to_images_script.py
kbab: Thanks - will remove relevant lines.

bia-ingest/tests/test_conversion_author_functions.py
kbab: Thanks - implemented as suggested

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,255 @@
...
+            )
+        specimen = specimen[0]
+
+        biosample_title = association["Biosample"]
+        biosample = [
+            f for f in filter(_find_element_by_title(biosample_title), biosample_dicts)
+        ]
+        if len(biosample) > 1:
+            logger.warning(
+                f"Submission has more than one biosample with the title {biosample_title}. Using first biosample for association {association}"
```
liviuba, line 105: ok now I get the issue with check-then-create for the api

But since this uses the api client directly instead of core, blocks like this (and below) can just be 
```
biosample_model = dicts_to_api_models([biosample,], Biosample)[0]
api_client.create_biosample(biosample_model, overwrite_mode=OverwriteMode.ALLOW_IDEMPOTENT)
```
with a single clear(er) code path

The catch with this however is that it will fail if the biosample (or any object) exists but was updated already (so it has version!=1) but that's ok since we never have that case, and if it would happen it would raise the issue of how do we merge different versions of the same object

+overwrite_mode gets typed as Any by openapi-codegen, will check tomorrow

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,255 @@
...
+    generate_specimen_uuid,
+    generate_image_acquisition_uuid,
+    extract_biosample_dicts,
+    extract_specimen_dicts,
+    extract_image_acquisition_dicts,
+    dicts_to_api_models,
+    dict_to_uuid,
+)
+
+logger = logging.getLogger(__file__)
```
liviuba, line 39: I'm not sure what's the intended use for this and bia_ingest/scli, but is this the same as the rw_client there (so that could be used just to have a single source of truth for how clients get created)?

bia-ingest/scripts/assign_image_acquisitions_to_images.py
```diff
@@ -0,0 +1,255 @@
...
+    )
+    return fileref_to_image_acquisitions
+
+
+def update_images_with_image_acquisitions(
+    images: List[BIAImage], fileref_to_image_acquisitions: Dict[str, List[str]]
+) -> None:
+    # Types of image representations that represent abstract BIAImage
+    bia_image_types = [
+        "fire_object",
```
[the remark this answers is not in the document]
kbab: Yes - it is. Thanks for the simpler version :)

bia-ingest/scripts/assign_image_acquisitions_to_images.py
kbab: Your suggestion is clearer and I will use it. However, I put the check for at least 1 image_acquisition_uuid so that we do not overwrite existing image_acquisition_uuids with empty lists. What do you think?

ctr26:
A lot of the data class code feels the same or similar to what bia-integrator has. Is this observation correct? Do we need to address this somewhere?

bia-ingest/pyproject.toml
liviuba: oh! I was just overjoyed we dropped core. We have https://pypi.org/project/bia-integrator-api/ and it would be nice to have "an internal client of a public thing" to make sure we ship something useful

There might be a catch because bia-ingest is technically internal and we might want to use some unreleased/dev/patched client. Some of the projects have a separate dev/pyproject.toml to switch easily, just not all of them / it wasn't explicitly agreed that we'd do that

kbab: Thanks @ctr26  and @liviuba . Changed to use pypi package. Pinned at 0.3.0

bia-ingest/scripts/assign_image_acquisitions_to_images.py
kbab: Thanks - implemented as suggested

bia-ingest/scripts/assign_image_acquisitions_to_images.py
kbab: Thanks. Changed as suggested.

bia-ingest/bia_ingest/conversion.py
```diff
@@ -0,0 +1,392 @@
...
+from .models import Author, BIAStudy, Affiliation, Publication
+from .biostudies import Section, Submission, Attribute, attributes_to_dict
+
+# Import these from API models
+from bia_integrator_api.models.biosample import Biosample
+from bia_integrator_api.models.specimen import Specimen
+from bia_integrator_api.models.image_acquisition import ImageAcquisition
+
+
+def find_sections_recursive(section: Section, search_types: List[str], results=[]):
```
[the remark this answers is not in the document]
kbab: Thanks. Added type hints to most functions (except some in tests)

bia-ingest/bia_ingest/biostudies.py
```diff
@@ -0,0 +1,325 @@
...
+    temp_key = key.lower()
+    if temp_key in keys:
+        key_index = keys.index(temp_key)
+        temp_key = list(dictionary.keys())[key_index]
+        return dictionary[temp_key]
+    else:
+        raise KeyError(f"{key} not in {dictionary.keys()}")
+
+
+def filter_filelist_content(dictionary: dict):
```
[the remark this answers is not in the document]
kbab: Good suggestion - but will revisit in future iteration as I did not write the original function

bia-ingest/bia_ingest/biostudies.py
```diff
@@ -0,0 +1,325 @@
...
+
+    return submission
+
+
+def attributes_to_dict(attributes: List[Attribute]) -> dict:
+
+    return {attr.name: attr.value for attr in attributes}
+
+
+def find_file_lists_in_section(section, flists) -> list:
```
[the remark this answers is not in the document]
kbab: Thanks - implemented

kbab:
> A lot of the data class code feels the same or similar to what bia-agent has. Is this observation correct? Do we need to address this somewhere?
> 
> As a wider comment, all of the the repos are getting more reliant on each other. The repos to me only serve to generate different CLIs. I'd be pro additional merging of repos and making a mega monolithic ?

Yes - we have similar models. @sherwoodf is taking a look at this and future iterations of the code base will attempt to address this. It will be good to discuss the structure of the repos at one of the team meetings.

PR #106, opened 2024-05-07
Remove api prefix

liviuba:
Changes needed to make the api work under the TM reverse proxy
* move openapi.json, docs out of root to versioned path
* remove /api/ prefix from root. This is so clients get generated with /vN/... paths and then we can configure clients to use https://ebi.ac.uk/bioimage-archive/api as the base (instead of https://ebi.ac.uk/bioimage-archive)

These are *very* breaking changes, all fixable with a client update (once the branch is merged in)

### DIFF
--- a/api/src/tests/util.py
+++ b/api/src/tests/util.py
@@ -324,7 +324,7 @@ def get_image_acquisition(
     return rsp.json()
 
 
-TEST_SERVER_BASE_URL = "http://localhost.com/api/v1"
+TEST_SERVER_BASE_URL = "http://localhost.com/v1"
 
 
 def get_client(**kwargs) -> TestClient:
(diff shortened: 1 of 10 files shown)


api/src/tests/util.py
```diff
@@ -324,7 +324,7 @@ def get_image_acquisition(
     return rsp.json()
 
 
-TEST_SERVER_BASE_URL = "http://localhost.com/api/v1"
+TEST_SERVER_BASE_URL = "http://localhost.com/v1"
```
sherwoodf, line 327: It almost feels like it would make the system behaviour clearer through the testing if the base URL stopped at .com/, and the v1 onwards was part of the path in the tests e.g. make_study made a post to v1/private/studies. I guess we might need to make that change if we ever did a version bump and wanted to test both versions, but this is unlikely.

sherwoodf (approved):

api/src/tests/util.py
liviuba: I agree, I think this was because versioning came after lots of code for the tests was written, so this was a quick solution to avoid a mass find-and-replace for every existing request in the tests

PR #108, opened 2024-05-17
added uuid type validation

sherwoodf:
Added logic to validate the type of an objected referenced by UUID conforms to the expected schema class. See https://app.clickup.com/t/869404wqc for more context. With this code the behaviour is:

Return 400 error for singular create & update of a document if any of it's dependent documents are not of the expected type.

Return  400 error in bulk create operations if: 
- any of it's dependent documents are not of the expected type
- across the bulk operations there is an expectation of more than 1 type for the same UUID 

Additionally, i had to split up the annotation tests to make the VSCode test plugin behave. See this ticket: https://app.clickup.com/t/8694k9nw9 for more context

### DIFF
--- a/api/src/tests/test_biosample_specimen_image_acquisition.py
+++ b/api/src/tests/test_biosample_specimen_image_acquisition.py
@@ -99,9 +99,9 @@ def test_image_acquisition_create_retrieve_update(
     assert rsp.status_code == 200, rsp.json()
 
     rsp = api_client.get(f"image_acquisitions/{uuid}")
-    specimen_fetched = rsp.json()
-    del specimen_fetched["model"]
-    assert specimen_fetched == image_acquisition
+    image_acquisition_fetched = rsp.json()
+    del image_acquisition_fetched["model"]
+    assert image_acquisition_fetched == image_acquisition
 
 
 def test_create_update_with_badly_typed_uuid(


api/src/tests/test_annotations_collection.py
```diff
@@ -0,0 +1,89 @@
+from src.tests.util import DBTestMixin, get_template_collection
```
sherwoodf, line 1: tests unchanged - i just had to split the original file up to get the tests working with VSCode

api/src/tests/test_annotations_fileref.py
```diff
@@ -0,0 +1,120 @@
+from src.models.repository import Repository
```
sherwoodf, line 1: tests unchanged - i just had to split the original file up to get the tests working with VSCode

api/src/tests/test_annotations_image.py
```diff
@@ -0,0 +1,115 @@
+from src.models.repository import Repository
```
sherwoodf, line 1: tests unchanged - i just had to split the original file up to get the tests working with VSCode

api/src/tests/test_annotations_study.py
```diff
@@ -0,0 +1,230 @@
+from src.models.repository import Repository
```
sherwoodf, line 1: tests unchanged - i just had to split the original file up to get the tests working with VSCode

api/src/models/repository.py
```diff
@@ -562,6 +524,103 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+            if len(dependency_info['type']) != 1:
+                doc = await self.biaint.find_one({"uuid": dependency_uuid})
+                for model_with_dependency_idx in dependency_info['index']:
+                    ref_bulk_operation_response.items[
+                        model_with_dependency_idx
+                    ].status = 400
+                    if doc is None:
+                        ref_bulk_operation_response.items[
+                            model_with_dependency_idx
+                        ].message = f"{dependency_uuid} does not exist. Additionally, your request expects it to be an instances of more than 1 conflicting types: {", ".join([x.__name__ for x in iter(dependency_info['type'])])}"
```
sherwoodf, line 571: Looks like i need to be more careful with my quotes to get the build working. Will fix once i've gotten comments (it was working locally so i don't know why it fails here)

api/src/models/repository.py
```diff
@@ -562,6 +524,103 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+                    else:
+                        ref_bulk_operation_response.items[
+                            model_with_dependency_idx
+                        ].message = f"Your request expects {dependency_uuid} to be an instance of more than 1 of conflicting types: {", ".join([x.__name__ for x in iter(dependency_info['type'])])}, when it is an instance of {doc["model"]["type_name"]}."
```
sherwoodf, line 575: Same as just above - will fix quites.

api/src/models/repository.py
```diff
@@ -562,6 +524,103 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+
+    async def validate_uuid_type(
+        self, target_uuid: uuid.UUID, expected_type: Type[models.DocumentMixin]
+    ):
+        doc = await self.biaint.find_one({"uuid": target_uuid})
+        if doc is None:
+            raise exceptions.DocumentNotFound(f"{target_uuid} does not exist")
+        if doc["model"]["type_name"] != str(expected_type.__name__):
+            raise exceptions.UnexpectedDocumentType(
+                f"{target_uuid} expected to be of type {expected_type.__name__}, but found {doc["model"]["type_name"]}"
```
kbab, line 601: Quote issue here as well

api/src/tests/test_file_reference.py
```diff
@@ -124,7 +124,7 @@ def test_create_file_references_multiple_errors(
         "E11000 duplicate key error" in bulk_write_results_by_status[400][7]["message"]
     )
     assert "Expected version to be 0" in bulk_write_results_by_status[400][5]["message"]
-    assert "Study does not exist" in bulk_write_results_by_status[400][3]["message"]
+    assert f"{file_references[3]["study_uuid"]} does not exist" in bulk_write_results_by_status[400][3]["message"]
```
kbab, line 127: quote issue here as well

api/src/tests/test_image.py
```diff
@@ -85,11 +92,16 @@ def test_create_images_multiple_errors(api_client: TestClient, existing_study: d
...
     }
 
+    assert "Expected version to be 0" in bulk_write_results_by_status[400][4]["message"]
     assert (
-        "E11000 duplicate key error" in bulk_write_results_by_status[400][7]["message"]
+        "E11000 duplicate key error" in bulk_write_results_by_status[400][5]["message"]
     )
-    assert "Expected version to be 0" in bulk_write_results_by_status[400][5]["message"]
-    assert "Study does not exist" in bulk_write_results_by_status[400][3]["message"]
+    assert f"{images[6]["study_uuid"]} does not exist" in bulk_write_results_by_status[400][6]["message"]
```
kbab, line 99: quote issue here. Also the start of lines 102 and 103 contain rogue single quote

api/src/tests/test_image.py
```diff
@@ -85,11 +92,16 @@ def test_create_images_multiple_errors(api_client: TestClient, existing_study: d
...
+    assert f"{images[7]["study_uuid"]} expected to be of type BIAStudy, but found Specimen" in bulk_write_results_by_status[400][7]["message"]
+    # Check that 2 different images using the same image_acquisition uuid for an image acquisition and a study both get an error message.
+    assert f"'Your request expects {images[9]["study_uuid"]} to be an instance of more than 1 of conflicting types: BIAStudy, ImageAcquisition, when it is an instance of ImageAcquisition." in bulk_write_results_by_status[400][8]["message"]
```
kbab, line 102: quote issue here. Also the start of lines 102 and 103 contain rogue single quote

api/src/tests/test_image.py
```diff
@@ -85,11 +92,16 @@ def test_create_images_multiple_errors(api_client: TestClient, existing_study: d
...
+    assert f"'Your request expects {images[9]["study_uuid"]} to be an instance of more than 1 of conflicting types: BIAStudy, ImageAcquisition, when it is an instance of ImageAcquisition." in bulk_write_results_by_status[400][9]["message"]
```
kbab, line 103: quote issue here. Also the start of lines 102 and 103 contain rogue single quote

api/src/models/repository.py
liviuba: +1 for the syntax errors, can't check the code without running it and it won't run because of that

kbab (commented):
Some code corrections needed. I was also confused by use of `specimen_fetched` in `test_image_acquisition_create_retrieve_update`  function.

api/src/models/repository.py
```diff
@@ -562,6 +524,119 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+        field_type_map: dict[str, Type[models.DocumentMixin]],
+    ):
+
+        for field, doc_type in field_type_map.items():
+
+            if isinstance(getattr(doc_to_verify, field), uuid.UUID):
+                uuid_to_check = getattr(doc_to_verify, field)
+                await self.validate_uuid_type(uuid_to_check, doc_type.__name__)
+
+            if isinstance(getattr(doc_to_verify, field), List):
```
liviuba, line 539: elif here would make it easier to tell that the two ifs exclude each other IMO

api/src/models/repository.py
```diff
@@ -562,6 +524,119 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+                        model_with_dependency_idx
+                    ].status = 400
+                    ref_bulk_operation_response.items[
+                        model_with_dependency_idx
+                    ].message = e.detail
+
+        return
+
+    async def validate_uuid_type(self, target_uuid: uuid.UUID, expected_type_name: str):
+        doc = await self.biaint.find_one({"uuid": target_uuid})
```
liviuba, line 600: we don't have really really large individual documents now, but maybe in the future we might (e.g. if we have lots of annotations, or save past versions of an object in the object itself)

might be good to be precise about what we pull since this sort of validation could easily get hidden away somewhere for lots of operations. Maybe something like find_one({...}, projection={'uuid': 1, 'model': 1})?

api/src/models/repository.py
```diff
@@ -562,6 +524,119 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+        if doc["model"]["type_name"] != expected_type_name:
+            raise exceptions.UnexpectedDocumentType(
+                f"{target_uuid} expected to be of type {expected_type_name}, but found {doc['model']['type_name']}"
+            )
+
+        return
+
+    def append_dependency(
+        self,
+        dependency_map: dict,
```
liviuba, line 612: Could this be static and the param named dependency_map_collector or something similar, to help suggest that the function does something non-obvious (collecting things in the argument instead of returning something)?

api/src/tests/test_biosample_specimen_image_acquisition.py
```diff
@@ -99,13 +99,64 @@ def test_image_acquisition_create_retrieve_update(
...
+
+def test_create_update_with_badly_typed_uuid(
+    api_client: TestClient,
+    uuid: str,
+    existing_specimen: dict,
+    existing_image_acquisition: dict,
+    existing_study: dict,
+):
+    image_acquisition = {
+        "uuid": uuid,
```
liviuba, line 115: The `get_template_*` functions are nicer for this IMO, because if we change something in the models we only have the template functions to change (learned this the hard way with the first model updates!)

So even if specimen_uuid needs to be different, I would rather just `acq = get_template_image_acquisition()` and then `acq["specimen_uuid"] = ...` (or just `get_template_image_acquisition(existing_study)`)

api/src/tests/test_image.py
```diff
@@ -50,13 +53,21 @@ def test_create_images_valid(api_client: TestClient, existing_study: dict):
...
-    images[5]["version"] = 2
+    images[4]["version"] = 2
     # mongo rejects *both* documents that violate an index constraint in a multi-doc insert
-    images[7]["uuid"] = images[0]["uuid"]
-    images[3]["study_uuid"] = "00000000-0000-0000-0000-000000000000"
+    images[5]["uuid"] = images[0]["uuid"]
+    images[6]["study_uuid"] = "00000000-0000-0000-0000-000000000000"
+    images[7]["study_uuid"] = existing_specimen["uuid"]
+    images[8]["image_acquisitions_uuid"] = [existing_image_acquisition["uuid"]]
+    images[9]["study_uuid"] = existing_image_acquisition["uuid"]
```
liviuba, line 70: 👍 really nice to check the dependency type here (and message) here too

api/src/models/repository.py
```diff
@@ -8,7 +8,7 @@
     AsyncIOMotorDatabase,
 )
 import pydantic
-from typing import List, Any, Callable
+from typing import List, Any, Callable, Type, Union
```
liviuba, line 11: Oh also the Callable here can be dropped since it's not used anymore

api/src/models/repository.py
```diff
@@ -562,6 +524,119 @@ async def get_specimen(self, *args, **kwargs) -> models.Specimen:
...
+        self,
+        doc_to_verify: models.DocumentMixin,
+        field_type_map: dict[str, Type[models.DocumentMixin]],
+    ):
+
+        for field, doc_type in field_type_map.items():
+
+            if isinstance(getattr(doc_to_verify, field), uuid.UUID):
+                uuid_to_check = getattr(doc_to_verify, field)
+                await self.validate_uuid_type(uuid_to_check, doc_type.__name__)
```
liviuba, line 537: Something to keep in mind if we ever do lazy db migrations. Some edge cases might pop up where we would need to extend this to check (model_type_name, version) pairs or similar, so we would propagate the lazy migration being applied to object dependencies

liviuba (approved):
Looks good! Few small comments but happy to merge.

What I'm worried about is the explicit check in the request handler, we might want to prioritise the ticket to do dependency checking automatically, especially if we're going to make changes to the models

PR #110, opened 2024-05-28
minor improvements to code readability

sherwoodf:
Minor fixes addressing comments from last PR:
https://github.com/BioImage-Archive/bia-integrator/pull/108

### DIFF
--- a/api/src/models/repository.py
+++ b/api/src/models/repository.py
@@ -8,7 +8,7 @@ from motor.motor_asyncio import (
     AsyncIOMotorDatabase,
 )
 import pydantic
-from typing import List, Any, Callable, Type, Union
+from typing import List, Any, Type, Union
 from enum import Enum
 import uuid
 import pymongo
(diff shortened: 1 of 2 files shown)


liviuba (commented):
Looks good! One thing I'd change is to leave uuid (the string) being passed in as a dependency so that tests that only create something look similar to the other create-related tests.

Both work, but conceptually, testing that a thing can be created IMO should assume everything else is assembled already (so you have the inidividual fields as primitive types, etc, passed in as dependencies).

If the test tests something more complex, then I would call get_template_X(add_uuid=True) just to get it out of the way since the uuid of the thing that's created isn't really so important that it's part of the test set-up

LE: Just realised but this wasn't explicitly stated. Since it's neater to split test parts into setup/test/teardown (except we don't have teardown), I tried to think of injected dependencies as the setup (or a maximal subset of it), just to have a slightly clearer delimitation between what's the setup and what's tested. This is why uuid being passed in for create-related tests was the default

liviuba (approved):
Looks good, happy to merge. There's a small issue to do with tests looking similar, do you want to change that before merging?

liviuba (approved):
LGTM!

PR #111, opened 2024-05-30
added initial version of bia shared datamodels

sherwoodf:
Example of linkml schema and generated pydantic classes for a study and related classes.


bia-shared-datamodels/src/pydantic/models.py
```diff
@@ -0,0 +1,121 @@
+from __future__ import annotations 
+from datetime import (
+    datetime,
+    date
+)
+from decimal import Decimal 
+from enum import Enum 
+import re
```
sherwoodf, line 8: Many of these imports don't actually get used.

bia-shared-datamodels/src/pydantic/models.py
```diff
@@ -0,0 +1,297 @@
...
+        description="""Number of files associated with the study."""
+    )
+    image_count: int = Field(
+        description="""Number of images associated with the study."""
+    )
+    see_also: Optional[List[AnyUrl]] = Field(
+        default_factory=list,
+        description="""Links to publications, github repositories, and other pages related to this Study.""",
+    )
+    fundedBy: Optional[Grant] = Field(
```
kbab, line 115: Should this be a list of grants?

bia-shared-datamodels/src/pydantic/models.py
```diff
@@ -0,0 +1,297 @@
...
+    A resource that acts or has the power to act.
+    """
+
+    display_name: str = Field(
+        description="""Name as it should be displayed on the BioImage Archive."""
+    )
+    contact_email: EmailStr = Field(
+        description="""An email address to contact the person or organisation"""
+    )
+    memberOf: Optional[List[Organisation]] = Field(
```
kbab, line 30: +1 - fixes https://app.clickup.com/t/8694enqx0

bia-shared-datamodels/src/bia_models/models.py
```diff
@@ -0,0 +1,343 @@
...
+    The manifestation of an image, represented in a particular image file format.
+    """
+
+    physical_dimension: Optional[str] = Field(
+        None, description="""Size of the physical space captured by the image"""
+    )
+    digital_dimension: Optional[str] = Field(
+        None, description="""Size in pixels or voxels of the image"""
+    )
+    image_viewer_setting: Optional[RenderedView] = Field(
```
sherwoodf, line 302: Should be list

bia-shared-datamodels/src/bia_models/models.py
```diff
@@ -0,0 +1,343 @@
...
+
+class ImageRepresentation(FileRepresentation):
+    """
+    The manifestation of an image, represented in a particular image file format.
+    """
+
+    physical_dimension: Optional[str] = Field(
+        None, description="""Size of the physical space captured by the image"""
+    )
+    digital_dimension: Optional[str] = Field(
```
sherwoodf, line 299: Should be required?

bia-shared-datamodels/src/bia_models/api_models.py
```diff
@@ -0,0 +1,295 @@
+from .semantic_models import (
```
sherwoodf, line 1: This file hasn't been updated in this commit, so is currently broken.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,608 @@
...
+    author: List[Contributor] = Field(description="""The creators of the document.""")
+    title: str = Field(
+        description="""The title of a scientific document. This will usually be displayed when search results including your data are shown."""
+    )
+    release_date: date = Field(description="""Date of first publication""")
+    keyword: Optional[List[str]] = Field(
+        default_factory=list,
+        description="""Keywords or tags used to describe the subject of a document""",
+    )
+    acknowledgement: Optional[List[Contributor]] = Field(
```
sherwoodf, line 85: Unsure of how we want to handle acknowledgements...

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,608 @@
...
+    """
+    An organsiation that a contributor is affiliated with.
+    """
+
+    display_name: str = Field(
+        description="""Name as it should be displayed on the BioImage Archive."""
+    )
+
+
+class DocumentMixin(BaseModel):
```
sherwoodf, line 71: The fields here make sense with a publication, but since publications aren't  our database's priortiy we might want to move some (e.g. keyword) to the study.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,608 @@
...
+        default_factory=list,
+        description="""Links to publications, github repositories, and other pages related to this Study.""",
+    )
+    related_publication: Optional[List[Publication]] = Field(
+        description="""The publications that the work involved in the study contributed to."""
+    )
+    grant: Optional[List[Grant]] = Field(
+        default_factory=list, description="""The grants that funded the study."""
+    )
+    funding_statement: Optional[str] = Field(
```
sherwoodf, line 118: I think we might want to nest funding statement & grant together on the website, but i'm not sure we need to do that here.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,608 @@
...
+        default_factory=list, description="""Description of how the study was funded."""
+    )
+    experimental_imaging_component: Optional[List[ExperimentalImagingDataset]] = Field(
+        default_factory=list,
+        description="""A dataset of that is associated with the study.""",
+    )
+    annotation_component: Optional[List[ImageAnnotationDataset]] = Field(
+        default_factory=list, description=""""""
+    )
+    attribute: dict = Field(
```
sherwoodf, line 128: We may want a mixin to inherit the attribute field.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,608 @@
...
+    acquisition_method: list[ImageAcquisition] = Field(
+        description="""Processes involved in the creation of the images and files in this dataset."""
+    )
+    specimen_preparation_method: list[SpecimenPrepartionProtocol] = Field(
+        description="""Processes involved in the creation of the samples that were then imaged."""
+    )
+    biological_entity: list[Biosample] = Field(
+        description="""The biological entity that was imaged."""
+    )
+    analysis_method: Optional[list[ImageAnalysisMethod]] = Field(
```
sherwoodf, line 364: might rename to acquisition protocol

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,608 @@
...
+    )
+    example_image_uri: list[str] = Field(
+        description="A viewable image that is typical of the dataset."
+    )
+    image_count: int = Field(
+        description="""Number of images associated with the dataset."""
+    )
+
+
+class ExperimentallyCapturedImage(AbstractImageMixin):
```
sherwoodf, line 378: We probably want a field to connect back from the image to the dataset it is from.
Also dataset->study?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    release_date: date = Field(description="""Date of first publication""")
+    keyword: Optional[List[str]] = Field(
+        default_factory=list,
+        description="""Keywords or tags used to describe the subject of a document""",
+    )
+    acknowledgement: Optional[List[Contributor]] = Field(
+        default_factory=list,
+        description="""Any person or group that should be acknowledged with the document.""",
+    )
+    description: Optional[str] = Field(
```
kbab, line 89: I think description is mandatory for studies in biostudies. Just mentioning, because I don't know if it is important for it to be mandatory for BIA as well.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+
+    accession_id: str = Field(description="""Unique ID provided by BioStudies.""")
+    license: LicenseType = Field(
+        description="""The license under which the data associated with the study is made avaliable."""
+    )
+    see_also: Optional[List[ExternalReference]] = Field(
+        default_factory=list,
+        description="""Links to publications, github repositories, and other pages related to this Study.""",
+    )
+    related_publication: Optional[List[Publication]] = Field(
```
kbab, line 112: Does this need a `default_factory=list,` to prevent the need of specifying an empty list if there are no related publications?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    A person or group that contributed to the creation of a Document.
+    """
+
+    display_name: str = Field(
+        description="""Name as it should be displayed on the BioImage Archive."""
+    )
+    contact_email: EmailStr = Field(
+        description="""An email address to contact the Contributor."""
+    )
+    role: str = Field(description="""The role of the contributor.""")
```
kbab, line 54: I think 'role' is optional in UI

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+
+class Contributor(PersonMixin, OrganisationMixin):
+    """
+    A person or group that contributed to the creation of a Document.
+    """
+
+    display_name: str = Field(
+        description="""Name as it should be displayed on the BioImage Archive."""
+    )
+    contact_email: EmailStr = Field(
```
kbab, line 51: Email is optional when creating 'contact' in UI (which translates to 'author')

bia-shared-datamodels/src/bia_models/semantic_models.py
kbab: Acknowledgements in the UI are simply a free text field with no structured information. In its current form it is not possible to programmatically parse it into a list of contributors

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    A published paper or written work.
+    """
+
+    pubmed_id: Optional[str] = Field(
+        None, description="""Identifier for journal articles/abstracts in PubMed"""
+    )
+    doi: str = Field(description="""Digital Object Identifier (DOI)""")
+
+
+class ExternalReference(BaseModel):
```
kbab, line 144: It will be interesting to see what happens with url validation here - UI does not enforce entering a url. It also allows a special prefix:ID format with it translates for known resources like biosamples etc. We may have to revisit this.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+        description="""A unique identifier for the grant, such as an Open Funder Registry ID.""",
+    )
+
+    funder: List[FundingBody] = Field(
+        default_factory=list,
+        description="""The name of the funding body providing support for the grant.""",
+    )
+
+
+class FundingBody(BaseModel):
```
kbab, line 174: The submission UI has a third unnamed column for 'fundings'. I can envision some users putting some sort of descriptive text in it. Should we have an optional 'description' column here? Or in UI remove the third column?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+class Grant(BaseModel):
+    """ """
+
+    id: Optional[str] = Field(
+        None,
+        description="""A unique identifier for the grant, such as an Open Funder Registry ID.""",
+    )
+
+    funder: List[FundingBody] = Field(
+        default_factory=list,
```
kbab, line 169: Does having a default of empty list here work? I assume since it  is not optional, the list will have to contain at least one entry?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    )
+
+
+class FileReference(BaseModel):
+    """
+    Information about a file, provided in file list.
+    """
+
+    file_name: str = Field(description="""The name of the file.""")
+    format: str = Field(description="""File format or type.""")
```
kbab, line 220: I am not sure what this field is for. We have 'type' already ...

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    size_in_bytes: int = Field(description="""Disc size in bytes.""")
+    uri: str = Field(description="""URI from which the file can be accessed.""")
```
kbab, line 222: Should this field be one of the url types?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    )
+    subject: Specimen = Field(
+        description="""The specimen that was prepared for and captured in the field of view of the image."""
+    )
+    submission_dataset: ExperimentalImagingDataset = Field(
+        description="""The dataset in which image was first submitted to the BIA."""
+    )
+
+
+class ImageAcquisition(BaseModel):
```
kbab, line 397: The UI has a field called 'title' in this section. Do we include this in 'imaging_instrument_description'?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    )
+    image_acquistion_parameters: str = Field(
+        description="""Parameters relevant to how the image was taken, such as instrument settings."""
+    )
+    fbbi_id: List[str] = Field(
+        description="""Biological Imaging Methods Ontology id indicating the kind of imaging that was perfomed."""
+    )
+
+
+class SpecimenPrepartionProtocol(BaseModel):
```
kbab, line 413: Where do we put 'title' from UI?

kbab (commented):
Looks good in general. I need to look at the google doc a bit more to review the artefacts pertaining to images and the different ways of modelling them.

Also I am a bit lost on the distinction between bia_data_model and semantic_models and which should I use and when. 

Does pyproject.toml need bia_models to be installed? - I needed 'from . import semantic_models' in bia_data_model.py

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: I've not found a uri library with a uri type for python 3.11

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Made optional

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: made optional

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+    """
+
+    display_name: str = Field(
+        description="""Name as it should be displayed on the BioImage Archive."""
+    )
+    contact_email: EmailStr = Field(
+        description="""An email address to contact the Contributor."""
+    )
+    role: str = Field(description="""The role of the contributor.""")
+    affiliation: Optional[List[Affiliation]] = Field(
```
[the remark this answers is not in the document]
sherwoodf: Made mandatory

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: switched to a string.

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Leaving as optional - since i suspect it;s optional for publications.

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: added. Note i'm not sure it needs this, but that's what linkml generates & i wanted to maintain consistency across all definitions so it's less confusing.

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Switched to optional.

I'm not actually sure that's what default list does though - just provides a default creation method of list, but i think it will create an empty list. I'm not sure if that then passes validation since the list will be empty, but it's what linkml generates so i kept it as 'standard'.

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Is that funding statement on studies, or is that something different?

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Added a 'user provided title/id' to the bia data model file.

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Added a 'user provided title/id' to the bia data model file.

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,619 @@
...
+
+    sample_of: List[Biosample] = Field(
+        description="""The biological matter that sampled to create the specimen."""
+    )
+    preparation_method: List[SpecimenPrepartionProtocol] = Field(
+        description="""How the biosample was prepared for imaging."""
+    )
+
+
+class Biosample(BaseModel):
```
sherwoodf, line 448: Added a 'user provided title/id' to the bia data model file.

sherwoodf:
> Also I am a bit lost on the distinction between bia_data_model and semantic_models and which should I use and when.

semantic_models is supposed to mirror the diagram from the docs - so it always points to-and-from classes that we've defined. bia_data_model takes those classes and provides a substitute for the fields where use some ID. I suspect we will be using bia_data_models 99% of the time, since it provides classes for the root of all the documents we want to create.

This of course is very much 'what i intended' with the split, and it may not pan out to work nicely in that way. If so we should definitely revisit the models to make them work better for us.

> Does pyproject.toml need bia_models to be installed? - I needed 'from . import semantic_models' in bia_data_model.py

import semantic_models is in the top of bia_data_model. The connections seem to work for me - have you done a poetry lock & poetry update --sync?

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    original_file_reference: Optional[List[UUID]] = Field()
+
+
+class ExperimentalImagingDataset(
+    semantic_models.ImageRepresentation,
+    DocumentMixin,
+    UserIdentifiedObject,
+):
+    image: List[UUID] = Field()
+    file: List[UUID] = Field()
```
kbab, line 49: Should this be `file_uri` to override `semantic_models.ImageRepresentation.file_uri` ?

sherwoodf: No - this is overriding the DatasetMixin field 'file' which points to a file representation, not the actual files.

kbab: Okay, but I am struggling to get to DatasetMixin from bia_data_model.ExperimentalImagingDataset. What am I missing?

sherwoodf: That's because i've inherited the wrong class!

sherwoodf: I've just pushed a new revision with fixes

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class ImageRepresentation(
+    semantic_models.ImageRepresentation,
+    DocumentMixin,
+):
+    original_file_reference: Optional[List[UUID]] = Field()
+
+
+class ExperimentalImagingDataset(
```
kbab, line 43: The parent class in semantic_models has fields referencing some objects that may need to be overridden here to allow the use of UUIDS. E.g. acquisition_method, specimen_preparation_method, biological_entity, etc.

kbab (commented):
I have not yet tried creating a Dataset (ExperimentalImageDataset or ImageAnnotationDataset), however, some of the attributes of the classes seem circular to me: e.g.

ExperimentalImageDataset.file is a list of FileReference/FileReferenceUUIDs, however, to create a FileReference you need a submission_dataset uuid . Do you think this may be an issue when creating objects?

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    submitted_in_study: UUID = Field()
+    # we include image analysis and correlation
+
+
+class ExperimentallyCapturedImage(
+    semantic_models.ExperimentallyCapturedImage,
+    DocumentMixin,
+):
+    acquisition_process: List[UUID] = Field()
+    represenatation: List[UUID] = Field()
```
kbab, line 59: representation?

bia-shared-datamodels/src/bia_models/semantic_models.py
```diff
@@ -0,0 +1,621 @@
...
+# Subgraph 4: Abstract images & their representations
+#######################################################################################################
+
+
+class AbstractImageMixin(BaseModel):
+    """
+    The abstract notion of an image that can have many representions in different image formats.
+    """
+
+    represenatation: List[ImageRepresentation] = Field(
```
kbab, line 250: representation?

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    DocumentMixin,
+    UserIdentifiedObject,
+):
+    image: List[UUID] = Field()
+    file: List[UUID] = Field()
+    submitted_in_study: UUID = Field()
+    # we include image analysis and correlation
+
+
+class ExperimentallyCapturedImage(
```
kbab, line 54: Do 'subject' and 'submission_dataset' from semantic_models need to be overwritten here?

kbab (commented):
Do we need DerivedImage in bia_data_model such that we can create some with UUIDs for the ImageAnnotationDataset?

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    semantic_models.AnnotationFileReference,
+    DocumentMixin,
+):
+    represenatation: List[UUID] = Field()
+    source_image: UUID = Field()
+    submission_dataset: UUID = Field()
+    creation_process: UUID = Field()
+
+
+class AnnotationMethod(
```
kbab, line 115: The should the "source_dataset" in semantic_models.AnnotationMethod be overridden here to allow use of UUIDs of ExperimentalImageDataset?

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+class SpecimenPrepartionProtocol(
+    semantic_models.SpecimenPrepartionProtocol,
+    DocumentMixin,
+    UserIdentifiedObject,
+):
+    pass
+
+
+class Specimen(
+    semantic_models.Specimen,
```
kbab, line 80: Are the UUID and title_id left out on purpose?

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    submitted_in_study: UUID = Field()
+    annotation_method: UUID = Field()
+
+
+class AnnotationFileReference(
+    semantic_models.AnnotationFileReference,
+    DocumentMixin,
+):
+    represenatation: List[UUID] = Field()
+    source_image: UUID = Field()
```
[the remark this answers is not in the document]
sherwoodf: Fixed

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    file: List[UUID] = Field()
+    submitted_in_study: UUID = Field()
+    annotation_method: UUID = Field()
+
+
+class AnnotationFileReference(
+    semantic_models.AnnotationFileReference,
+    DocumentMixin,
+):
+    represenatation: List[UUID] = Field()
```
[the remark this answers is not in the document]
sherwoodf: Not needed - i think i forgot to do derived image / combined the two here.

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+    pass
+
+
+class Specimen(
+    semantic_models.Specimen,
+):
+    sample_of: List[UUID] = Field()
+    preparation_method: List[UUID] = Field()
+
+
```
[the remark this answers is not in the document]
sherwoodf: This is following on from what we previously used for biosample - but other than that it's not really intetional.

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -0,0 +1,120 @@
...
+class Biosample(
+    semantic_models.Biosample,
+    DocumentMixin,
+    UserIdentifiedObject,
+):
+    pass
+
+
+class ImageAnnotationDataset(
```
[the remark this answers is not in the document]
sherwoodf: yes

bia-shared-datamodels/src/bia_models/bia_data_model.py
sherwoodf: ... Yes, though actually i think we might want to allow it to be a uri as well (so that people can link out to other datasets that they might have annotated)

bia-shared-datamodels/src/bia_models/semantic_models.py
sherwoodf: Yes - also need to fix in bia_data_models

bia-shared-datamodels/src/bia_models/bia_data_model.py
sherwoodf: submission_dataset - yes
subject - no, but ah, i've just realised this complicates things....

sherwoodf: Specimen needed to be overriden as a class - i don't think it's worth it having it's own uuid as it's currently just linking the preparation protocol + the biosample together - linking both of these directly from the image i think improves the usability of the api because you don't have to make so many requests one after another to get all the info on what an image is of (which is pretty important for the image)

bia-shared-datamodels/src/bia_models/bia_data_model.py
sherwoodf: Yes, because users aren't providing specimens (they provide specimen preparation protocols) and the specimens are now just links to biosamples + specimen preparation protocol, so i don't think it's worth them having their own uuids outside of the image...? maybe. Perhaps that needs more discussion... but it feels like by keeping it inside the image document we reduce the number of calls a user needs to make to get at the relevant data for an image and don't loose out on much.

bia-shared-datamodels/src/bia_models/bia_data_model.py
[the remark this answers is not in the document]
kbab: I can go with BioSample, but it seems quirky to me, as when I instantiate the class in python I would translate it to bio_sample - which looks odd compared to biosample...

kbab (approved):
LGTM

bia-shared-datamodels/src/pydantic/models.py
ctr26: I use 

`ruff --fix `

for removing used imports

ctr26 (approved):

PR #114, opened 2024-06-28
Shared models v2

kbab:
Add tests for creation of all models in bia_data_model and minor tweaks to the various model definitions

### DIFF
--- a/bia-shared-datamodels/src/bia_models/bia_data_model.py
+++ b/bia-shared-datamodels/src/bia_models/bia_data_model.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import semantic_models
-from pydantic import BaseModel, Field, AnyUrl
+from pydantic import BaseModel, Field, AnyUrl, conlist
 from typing import List, Optional, Union
 from uuid import UUID
 
@@ -57,8 +57,8 @@ class ExperimentalImagingDataset(
 
 
 class Specimen(semantic_models.Specimen):
-    preparation_method: List[UUID] = Field()
-    sample_of: List[UUID] = Field()
+    preparation_method: conlist(item_type=UUID, min_length=1) = Field()
+    sample_of: conlist(item_type=UUID, min_length=1) = Field()
 
 
 class ExperimentallyCapturedImage(
--- a/bia-shared-datamodels/src/bia_models/bia_data_model.py
+++ b/bia-shared-datamodels/src/bia_models/bia_data_model.py
@@ -26,6 +26,8 @@ class Study(
 ):
     experimental_imaging_component: List[UUID] = Field()
     annotation_component: List[UUID] = Field()
+    author: conlist(item_type=semantic_models.Contributor, min_length=1) = Field()
+    description: str = Field()
 
 
 class FileReference(
--- a/bia-shared-datamodels/test/test_shared_models.py
+++ b/bia-shared-datamodels/test/test_shared_models.py
@@ -1,0 +1,44 @@
+import pytest
+from . import utils
+from .utils import (
+    bia_data_model,
+    semantic_models,
+)
+
+
+@pytest.mark.parametrize(
+    ("expected_model_type", "model_creation_func",),
+    (
+        (bia_data_model.Study, utils.get_template_study,),
+        (bia_data_model.FileReference, utils.get_template_file_reference,),
+        (bia_data_model.ImageRepresentation, utils.get_template_image_representation,),
+        (
+            bia_data_model.ExperimentalImagingDataset,
+            utils.get_template_experimental_imaging_dataset,
+        ),
+        (bia_data_model.Specimen, utils.get_template_specimen,),
+        (
+            bia_data_model.ExperimentallyCapturedImage,
+            utils.get_template_experimentally_captured_image,
+        ),
+        (bia_data_model.ImageAcquisition, utils.get_template_image_acquisition,),
+        (
+            bia_data_model.SpecimenPrepartionProtocol,
+            utils.get_template_specimen_preparation_protocol,
+        ),
+        (bia_data_model.BioSample, utils.get_template_biosample,),
+        (
+            bia_data_model.ImageAnnotationDataset,
+            utils.get_template_image_annotation_dataset,
+        ),
+        (
+            bia_data_model.AnnotationFileReference,
+            utils.get_template_annotation_file_reference,
+        ),
+        (bia_data_model.DerivedImage, utils.get_template_derived_image,),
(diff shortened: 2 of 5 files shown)


kbab:
@sherwoodf  - some general comments:
* I would have defined semantic_models.ImageRepresentation.size_t as 'int' - I thought it was the size of the t dimension (number of timesteps), not the units
* Should semantic_models.DerivedImage.source_image point to a list of ExperimentallyCapturedImage as opposed to ImageRepresentation? (see semantic_model.AnnotationMixin)

sherwoodf:
> I would have defined semantic_models.ImageRepresentation.size_t as 'int' - I thought it was the size of the t dimension (number of timesteps), not the units
Ah that makes more sense - feel free to change that
> DerivedImage.source_image 
I thought our discussion was that image representation was necessary since a particular file format would have been annotated & transformations between those formats might mean annotations do not apply correctly to other representations

bia-shared-datamodels/test/test_shared_models.py
```diff
@@ -0,0 +1,44 @@
+import pytest
+from . import utils
+from .utils import (
+    bia_data_model,
+    semantic_models,
+)
+
+
+@pytest.mark.parametrize(
```
sherwoodf, line 9: I really like this testing method. We could then create conftest files in other pacakges that use these util methods to provide fixtures, but by testing them here it makes it super clear when models / data has changed.

bia-shared-datamodels/test/test_shared_models.py
```diff
@@ -0,0 +1,44 @@
...
+        ),
+        (
+            bia_data_model.AnnotationFileReference,
+            utils.get_template_annotation_file_reference,
+        ),
+        (bia_data_model.DerivedImage, utils.get_template_derived_image,),
+        (bia_data_model.AnnotationMethod, utils.get_template_annotation_method,),
+    ),
+)
+def test_create_models(expected_model_type, model_creation_func):
```
sherwoodf, line 42: Is there anything that would be good to test to make sure they fail to get created (and we get sensible error messages from those fails)? E.g. Specimen with empty lists? This might be overkill so i won't block this pr on it.

bia-shared-datamodels/test/utils.py
```diff
@@ -0,0 +1,351 @@
...
+    return biosample
+
+
+# Depends on:
+#   bia_data_model.BioSample
+#   bia_data_model.SpecimenPreparationProtocol
+def get_template_specimen() -> bia_data_model.Specimen:
+    specimen = bia_data_model.Specimen.model_validate(
+        {
+            "preparation_method": [get_template_specimen_preparation_protocol().uuid,],
```
sherwoodf, line 85: Whilst this makes it explicit what it depends on, would just calling uuid be simpler? I don't think we can do the actual uuid object type validation testing without setting up a DB to test.

bia-shared-datamodels/test/utils.py
```diff
@@ -0,0 +1,351 @@
...
+def get_template_derived_image() -> bia_data_model.DerivedImage:
+    derived_image = bia_data_model.DerivedImage.model_validate(
+        {
+            "uuid": uuid4(),
+            "source_image": [get_template_image_representation().uuid,],
+            "submission_dataset": get_template_image_annotation_dataset().uuid,
+            "creation_process": get_template_annotation_method().uuid,
+            "representation": [get_template_image_representation().uuid,],
+            "transformation_description": "Template transformation description",
+            "spatial_information": "Template spatial information",
```
sherwoodf, line 139: Ah i think: spatial_information, transformation_description, and creation_process probably need to be optional on images, as we will mostly get that at the dataset level only... I don't think it changes the test as written here though.

bia-shared-datamodels/test/utils.py
```diff
@@ -0,0 +1,351 @@
...
+            "file_uri": ["https://dummy.uri.org",],
+            "total_size_in_bytes": 0,
+            "physical_size_x": 1,
+            "physical_size_y": 1,
+            "physical_size_z": 1,
+            "size_x": 1,
+            "size_y": 1,
+            "size_z": 1,
+            "size_c": 1,
+            "size_t": 1,
```
sherwoodf, line 298: We might want to update the description of the size_t, as well as the type hint.

sherwoodf (approved):

bia-shared-datamodels/test/utils.py
```diff
@@ -0,0 +1,351 @@
...
+            "rorid": "None",
+            "address": "None",
+            "website": "https://www.none.com",
+            "orcid": "None",
+        }
+    )
+    return contributor
+
+
+def get_template_study() -> bia_data_model.Study:
```
sherwoodf, line 333: Oh i think we might be missing a description on study.

bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -57,8 +57,8 @@ class ExperimentalImagingDataset(
 
 
 class Specimen(semantic_models.Specimen):
-    preparation_method: List[UUID] = Field()
-    sample_of: List[UUID] = Field()
+    preparation_method: conlist(item_type=UUID, min_length=1) = Field()
+    sample_of: conlist(item_type=UUID, min_length=1) = Field()
```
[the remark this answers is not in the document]
kbab: Made Study.authors conlist and created [clickup task](https://app.clickup.com/t/8694y77c9) to discuss whether to apply to attributes of other models.

bia-shared-datamodels/test/test_shared_models.py
kbab: Done for Specimen

bia-shared-datamodels/test/utils.py
kbab: Yes you are correct. Being explicit helped me to reinforce my understanding of the dependencies between the models.

bia-shared-datamodels/test/utils.py
kbab: Done

bia-shared-datamodels/test/utils.py
kbab: Yes - thanks. They are both optional in semantic_models.DocumentMixin - added to test. Is it worth overriding description in bia_data_model.Study to make it mandatory? It is mandatory for actual submissions.

sherwoodf: Hmm. On reflection, from the point of view of displaying it on the website it might be useful to always have a description, so making it mandatory might be best....

PR #115, opened 2024-06-28
fix lack of inheritance to protocol mixin class

sherwoodf:
Fixed inheritance of the Image Acquisition and Specimen Preparation protocol: they were not inheriting the protocol mixin


bia-shared-datamodels/src/bia_models/bia_data_model.py
```diff
@@ -26,7 +26,7 @@ class Study(
 ):
     experimental_imaging_component: List[UUID] = Field()
     annotation_component: List[UUID] = Field()
-    author: conlist(item_type=semantic_models.Contributor, min_length=1) = Field()
+    author: semantic_models.Contributor = Field(min_length=1)
```
kbab, line 29: Does this need to be List[semantic_models.Contributor] = Field(min_length=1)

kbab (changes requested):
Only thing I spotted was overriding author in bia_data_model.Study. Looks good otherwise. Happy for merge after you have a look

bia-shared-datamodels/src/bia_models/bia_data_model.py
sherwoodf: yes!

PR #116, opened 2024-07-01
minor fixes to align annotation model, as well as licence spelling …

sherwoodf:
Creation process was a list on the annotation dataset (similar to image acquisition, to allow for splitting into multiple smaller steps). So we needed the DerivedImage and AnnotationFileReference to also have the list value to mirror the information on the dataset.

Also updated the spelling of license to licence.

An additional changes i'm considering, but have not included, is trying to further standardise when we put mini-count 1 restrictions on list fields, vs when we have optional list fields - suggestions welcome on this topic.


kbab (approved):
LGTM!

PR #118, opened 2024-07-03
Ingest from biostudies version 1

kbab:
First iteration of ingest using semantic_models and bia_data_models. We can now ingest studies into bia_data_model.Study. However, we do not get details of experimental imaging components or annotation components yet.


bia-ingest-shared-models/bia_ingest_sm/biostudies.py
sherwoodf, line 1: File/directory structure comments:

1. I would have expected this to be a copy of the existing biostudies.py file in the existing bia-ingest package, but i see a commented out block - is that the only change?
2. Presumably once this is the only ingest path we will remove bia-ingest and rename this bia-ingest (i.e. drop the -shared-models because it's redundant)
3. I was expecting these files to be in a src/ folder, rather than bia_ingest_sm.

bia-ingest-shared-models/bia_ingest_sm/cli.py
```diff
@@ -0,0 +1,23 @@
+import typer
+from typing import Optional
+from typing_extensions import Annotated
+from bia_ingest_sm.biostudies import load_submission
+from bia_ingest_sm.conversion import get_study
+
+app = typer.Typer()
+
+
+@app.command(help="Ingest from biostudies and echo json of bia_data_model.Study")
```
sherwoodf, line 10: Is the intent of this tool that it will eventually perform the api calls in order to ingest a study, or just create the json required for the submissiom?

bia-ingest-shared-models/bia_ingest_sm/conversion.py
```diff
@@ -0,0 +1,304 @@
...
+        "keyword": study_attributes.pop("Keywords", []),
+        "author": [c.model_dump() for c in contributors],
+        "grant": [g.model_dump() for g in grants],
+        "attribute": study_attributes,
+        "experimental_imaging_component": [],
+        "annotation_component": [],
+    }
+    study_uuid = dict_to_uuid(study_dict, ["accession_id",])
+    study_dict["uuid"] = study_uuid
+    study = bia_data_model.Study.model_validate(study_dict)
```
sherwoodf, line 47: still not used to model_validate as a way of instanciating pydantic classes but i guess it's the standard thing to be doing.

bia-ingest-shared-models/bia_ingest_sm/conversion.py
```diff
@@ -0,0 +1,304 @@
...
+        study_title = study_section_attr_dict.get("Title", "Unknown")
+
+    return study_title
+
+
+def get_licence(study_attributes: Dict[str, Any]) -> semantic_models.LicenceType:
+    """Return enum version of licence of study
+
+    """
+    licence = re.sub(r"\s", "_", study_attributes.get("License", "CC0"))
```
sherwoodf, line 68: Can the enum have spaces in it's value? that might just be easier?

bia-ingest-shared-models/test/test_shared_models.py
```diff
@@ -0,0 +1,27 @@
...
+        # urls
+        # (utils.get_test_external_reference, conversion.get_external_reference,),
+        # Do not test semantic_models.Publication yet. Need to resolve
+        # issues around some fields being mandatory or optional
+        # (utils.get_test_publication, conversion.get_publication,),
+        # (bia_data_model.Study, conversion.get_study_from_submission,),
+    ),
+)
+def test_create_models(expected_model_func, model_creation_func, test_submission):
+    expected = expected_model_func()
```
sherwoodf, line 25: i was shocked that this worked! Very cool though.

bia-ingest-shared-models/test/utils.py
```diff
@@ -0,0 +1,477 @@
+"""Utility functions to create models
```
sherwoodf, line 1: shouldn't we be re-using the ones defined in the-shared-models code?

bia-ingest-shared-models/bia_ingest_sm/biostudies.py
kbab: > File/directory structure comments:
> 
> 1. I would have expected this to be a copy of the existing biostudies.py file in the existing bia-ingest package, but i see a commented out block - is that the only change?

No - there are a few other changes, but they are very similar otherwise

> 2. Presumably once this is the only ingest path we will remove bia-ingest and rename this bia-ingest (i.e. drop the -shared-models because it's redundant)

Yes

> 3. I was expecting these files to be in a src/ folder, rather than bia_ingest_sm.

Happy to restructure - can do this with [2] above

bia-ingest-shared-models/bia_ingest_sm/cli.py
kbab: My understanding is that it will perform the API calls. Current version is quick and dirty to allow us to create json files for test exporter and new website

bia-ingest-shared-models/bia_ingest_sm/conversion.py
kbab: I agree, their website also has spaces in the name. Should I change this in semantic_models.py?

bia-ingest-shared-models/test/utils.py
kbab: Ideally we should. The tests in bia-shared-models only apply to 'bia_data_model' classes and the purpose was to create valid model instances.

Here we are testing classes from both semantic_models.py and bia_data_model.py. We also test that the created models align with expected values from a test submission. The tests here are more comprehensive, but we could move the whole test suite to bia-shared-models and just have one set of tests. Happy to do so in a further iteration.

bia-ingest-shared-models/bia_ingest_sm/conversion.py
sherwoodf: I think changing it makes sense - would we still want to do the regex to verify that the right kind of spaces are used?

bia-ingest-shared-models/test/utils.py
sherwoodf: Yeah, i guess the placeholder tests are about example models, whereas here they are to validate the output of conversion.

sherwoodf (approved):

PR #120, opened 2024-07-05
Ingest from biostudies version 2

kbab:
Create ExperiementalImagingDatasets, FileReferences, Biosample and ImageAnalysisMethod. Persist these to disk.

Because of time constraints the source code is sometimes verbose and needs refactoring. Additional notes:

* BioSample does not capture biological_entity

* FileReference.experimental_imaging_dataset.uuid is not yet implemented, current values are just copy of FileReference.uuid

* Persisting to disk messy (bia_data_dir env variable or .env file)

* Tested successfully on S-BIAD1026 and S-BIAD1250

* Failed on S-BIAD1260 (failed - zip files?) and  S-BIAD1040 (was taking too long > 100K files)

PR #121, opened 2024-07-05
added example input and output data

sherwoodf:
Manually created prototypical examples of study, dataset, image, representation, and file reference objects from existing studies and images.

There's likely to be some errors in this due to creating them manually, and they've highlighted certain differences between the old and new models which may need discussing, but these might be good to combine with kola's PR (https://github.com/BioImage-Archive/bia-integrator/pull/120) to test the conversion logic.


ctr26:
Is this branched from something else other than main? There's more file changes than I would expect

sherwoodf:
No, it's from main. It's 18 files because it's 4 studies & for each there is:
- Study page tab
- 1 image json from the API
- 1 file ref json from the api
- 1 output 'expected' json

And for 1 there is an additional annotation image + the file reference for that image.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -0,0 +1,189 @@
...
+            "UUID12-file_reference"
+        ],
+        "submitted_in_study": "UUID1-study",
+        "specimen_preparation_method": [
+            "UUID3-specimen_preparation_protocol"
+        ],
+        "acquisition_method": [
+            "UUID4-image-acquisition"
+        ],
+        "biological_entity": [
```
AybukeKY, line 56: This should be biosample not biological_entity. biological_entity is a field of/entry in biosample.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -0,0 +1,189 @@
...
+        "analysis_method": [],
+        "correlation_method": [],
+        "file_reference_count": 1,
+        "image_count": 1,
+        "example_image_uri": []
+    },
+    {
+        "uuid": "UUID3-specimen_preparation_protocol",
+        "title_id": "Pycnogonum litorale amputees",
+        "method_description": "All specimens were fixed and stored in Bouin’s fluid (10% formaldehyde, 5% glacial acetic acid in saturated aqueous picric acid) at room temperature (RT). Specimens were briefly rinsed in phosphate-buffered saline (PBS; 1.86 mM NaH2PO4, 8.41 mM Na2HPO4, 175 mM NaCl, pH 7.4; at least 3× 5 min), transferred into deionized water, dehydrated via an ascending ethanol series, incubated in a tissue contrasting solution of 2% iodine (resublimated; Carl Roth; #X864.1) in 99.5% ethanol for 48 h at RT, rinsed in 99.5% ethanol (3 to 4× 10 min), and critical point-dried with a Leica EM CPD300.",
```
AybukeKY, line 68: I'm confused about this. Specimen section in our input/submission models has two methods/protocols: "sample preparation protocol" which is mandatory and "growth protocol" which is optional. I think the name mapping here is confusing.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -0,0 +1,189 @@
...
+    },
+    {
+        "uuid": "UUID4-image-acquisition",
+        "title_id": "Pycnogonum litorale µCT",
+        "method_description": "MISSING - WE SHOULD WORK OUT WHERE TO GET THIS FROM",
+        "imaging_instrument_description": "Xradia MicroXCT-200 (Carl Zeiss Microscopy)",
+        "image_acquisition_parameters": "Dried specimens were placed in plastic tubes for overview scans. For higher resolution scans of the posterior body region with the regenerates, they were subsequently attached to plastic welding rods with hot glue. Scans were performed under 40 kV/200 μA/8 W or 30 kV/200 μA/6 W settings. Depending on the specimen size and region of interest, a 4×, 10×, or 20× objective was chosen. Exposure times were individually adjusted for each scan, ranging from 0.75 to 6.5 s. To reduce noise, binning 2 was applied during data acquisition. Tomography projections were reconstructed with the XMReconstructor software (Carl Zeiss Microscopy) with binning 1 (= full resolution) and TIFF format image stacks as output.",
+        "fbbi_id": [
+            "http://purl.obolibrary.org/obo/FBbi_00001002"
```
AybukeKY, line 77: This webpage should be working but it isn't currently :/

AybukeKY: Isn't method_description same as imaging_method (in the input json or corresponding fbbi id)?

bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -0,0 +1,189 @@
...
+    },
+    {
+        "uuid": "UUID3-specimen_preparation_protocol",
+        "title_id": "Pycnogonum litorale amputees",
+        "method_description": "All specimens were fixed and stored in Bouin’s fluid (10% formaldehyde, 5% glacial acetic acid in saturated aqueous picric acid) at room temperature (RT). Specimens were briefly rinsed in phosphate-buffered saline (PBS; 1.86 mM NaH2PO4, 8.41 mM Na2HPO4, 175 mM NaCl, pH 7.4; at least 3× 5 min), transferred into deionized water, dehydrated via an ascending ethanol series, incubated in a tissue contrasting solution of 2% iodine (resublimated; Carl Roth; #X864.1) in 99.5% ethanol for 48 h at RT, rinsed in 99.5% ethanol (3 to 4× 10 min), and critical point-dried with a Leica EM CPD300.",
+    },
+    {
+        "uuid": "UUID4-image-acquisition",
+        "title_id": "Pycnogonum litorale µCT",
+        "method_description": "MISSING - WE SHOULD WORK OUT WHERE TO GET THIS FROM",
```
AybukeKY, line 73: Isn't method_description same as imaging_method (in the input json or corresponding fbbi id)?

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: Oh I see now you've changed it in the models...

bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -0,0 +1,189 @@
...
+            ]
+        },
+        "attribute": {}
+    },
+    {
+        "uuid": "UUID7-image_representation",
+        "original_file_reference": [
+            "UUID12-file_reference"
+        ],
+        "image_format": "fire_objectt",
```
AybukeKY, line 119: fire_object

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: Imaging method currently seems to be just the FBBI ID / name of the entity in FBBI id. Judging by what i've seen, image_acquisition_parameters actually has this information each time. E.g. this one talks about about plastic tubes and noise reduction etc. not just the settings of the the method.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: Works now!

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: I do wonder if we want to change this to be the actual type of the file, or rename the field because i'm not sure the two match up currently. Depends on how we expect to use this field.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: Yep. The biological_entity field was basically always used as a description because the description field was optional. I suspect in situations where we have both concatenating them is the best option?

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: As in by having the generic 'method_description' and the more specific 'growth protocol description' it's less clear where to put growth protocol information because it could be in both?

AybukeKY: Yes and naming the whole class/section as 'specimen preparation protocol' which makes sense but can also be easily confused with 'sample preparation protocol'. I feel like we need to be a bit more reserved in changing the names that came out of bio-imaging community meetings aka REMBI.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: The plastic tubes etc isn't imaging method description in my mind as imaging method is completely different, e.g. X-ray computed tomography in this case. Maybe we can check the Quarep-Limi list and see if there's something more fitting to what this field is aiming for. Or we can change the "image acquisition parameters" to "image acquisition details" to cover all?

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: Still not working for me; tried both Safari and Chrome. Maybe it's my settings?...

bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -0,0 +1,189 @@
...
+        "total_size_in_bytes": 48796,
+        "size_x": 256,
+        "size_y": 256,
+        "image_viewer_setting": [],
+        "attribute": {}
+    },
+    {
+        "uuid": "UUID11-image_representation",
+        "original_file_reference": [],
+        "image_format": "ome_ngff",
```
AybukeKY, line 165: This isn't ome_ngff. I think it should be "representative".

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: Oh it's stopped working for me now....

bia-ingest-shared-models/test/data/S-BIAD634/output.json
```diff
@@ -0,0 +1,514 @@
+[
+    {
+        "uuid": "UUID-study-1",
```
AybukeKY, line 3: Do we want to use the actual study uuid's that are already in the API or not to mix them since we don't have for biosample etc.?

sherwoodf: I was going with not real uuids since we don't have them for biosamples etc. But i'm starting to understand better how the UUIDs are generated, so i would be more okay to switching to the real thing since i think the UUIDs can be generated regardless of the API (since they're based on specific field values)

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: Protocol description would be much better I think as for instance imaging method is a reserved word in the fbbi ontology.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: Well i'm partly quite keen to have a different name that includes protocol because it isn't the same concept that's in REMBI (because, as you've said, we can't get users to provide information on an image by image basis, which is what REMBI attempts to model).

I'm still quite keen to keep description on string descriptions to distinguish them from ontology IDs & not get the two confused - that will keep the field space open for more options.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: Yeah I know :/

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: But if you think it would be better for the biosample to have biological_entity_description then that makes sense. I jsut wanted two fields on specimen that make the difference clear without explicitly re-using the class names where possible, but this is probably a case where it doesn't matter much.

bia-ingest-shared-models/test/data/S-BIAD634/output.json
```diff
@@ -0,0 +1,514 @@
...
+        "attribute": {}
+    },
+    {
+        "uuid": "UUID-file_reference-1",
+        "file_name": "dataset/rawimages/normal_20.tif",
+        "format": "fire_object",
+        "size_in_bytes": 1311278,
+        "uri": "https://ftp.ebi.ac.uk/biostudies/fire/S-BIAD/606/S-BIAD606/Files/PycLit_InsVI_3_postMolt6_posBody.tif",
+        "attribute": {
+            "file description": "Raw nuclear images in TIFF format",
```
AybukeKY, line 426: We had the idea of having source of information for attributes, something like e.g. "file list", "image headers", "annotators". It may be a good time to try to implement that as well?

bia-ingest-shared-models/test/data/S-BIAD634/output.json
```diff
@@ -0,0 +1,514 @@
...
+        "annotation_criteria": "The annotation of nuclei in tissue sections or tumor touch imprints is challenging and may not be unambiguous due to out-of-focus light or nuclei, damaged nuclei or nuclei presenting with modified morphology due to the slide preparation procedure. We defined the following criteria to annotate nuclear images: Only intact nuclei are annotated, even if the nuclear intensity is low in comparison to all other nuclei present. Nuclei have to be in focus. If parts of a nucleus are out of focus, only the part of the nucleus being in focus is annotated. Nuclear borders have to be annotated as exact as resolution and blurring allows for. Nuclei are not annotated if their morphology was heavily changed due to the preparation procedure. Nuclei from dividing cells are annotated as one nucleus unless clear borders can be distinguished between the resulting new nuclei."
+    },
+    {
+        "uuid": "UUID-derived-image-1",
+        "source_image": [],
```
AybukeKY, line 467: source_image should be "dataset/rawimages/normal_20.tif". It's in the attributes of the file reference but it should be here as well. Is it because the name of the column is different in the file list? Note that we've very recently changed it to be source_image and not doing any checks so we need a fuzzy-match/catch-all case during ingest and include the source_image value here.

sherwoodf: Yes, it should be filled in, though actually should be UUID-image_representation-1 according to the model i think.,

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: Yes because in this case it is in the description of the biosample but in the other examples in the PR, it's actually lost because there's a separate description field in addition to biological_entity.

bia-ingest-shared-models/test/data/S-BIAD606/output.json
sherwoodf: It works today! XD

bia-ingest-shared-models/test/data/S-BIAD606/output.json
AybukeKY: I'd actually say the opposite for REMBI; it's a study and dataset based model rather than image by image which is what we're trying to do while trying to map to REMBI which is used by our submission form.

AybukeKY: To note the outcome of our discussion: growth protocol to be separated from sample preparation protocol, and linked to specimen.

bia-ingest-shared-models/test/data/S-BIAD634/output.json
sherwoodf: Will consider in other task / ticket

PR #122, opened 2024-07-09
Minor ingest logic refactor

sherwoodf:
I will be adding more conversion logic for converting different objects (image acquisition, specimen protocols etc)

Before doing so I thought it best to break up the conversion logic into separate files for each object we convert to, rather than work on a single megafile for conversion logic.


liviuba:
LGTM!

liviuba (approved):

PR #125, opened 2024-07-23
created bia-export base package

sherwoodf:
Set up the basics of the bia-export package to create studies for website from json files

### DIFF
--- a/bia-export/README.md
+++ b/bia-export/README.md
@@ -1,0 +1,32 @@
+BIA Export
+==========
+
+Export data from the BIA to feed static pages, and other downstream consumers. This:
+
+* Pulls data from the BIA Integrator API
+* Selects attributes for images and studies
+* Derives information from OME-Zarr representations (physical dimensions, axis sizes)
+* Transforms to a specific export format
+* Writes the result to a JSON file
+
+Note that the export process caches results, so be careful!
+ 
+Installation
+------------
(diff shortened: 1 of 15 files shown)


bia-export/README.md
```diff
@@ -0,0 +1,32 @@
+BIA Export
```
sherwoodf, line 1: I have not updated the readme - this information is not correct

bia-export/bia_export/website_conversion.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import List
+
+logger = logging.getLogger(__name__)
+
+def create_study(
+        accession_id: str,
+        root_directory: Path
+) -> Study:
+    if root_directory:
+        study_path = root_directory.joinpath(f'studies/{accession_id}.json')
```
sherwoodf, line 19: Note - these paths are based on the cache path structure of the ingest code. The idea is we should be able to run the export with the root based on the local 'cache' after running ingest and get the json for the website

bia-export/test/conftest.py
```diff
@@ -0,0 +1 @@
+import pytest
```
sherwoodf, line 1: Ah i should have probably removed this - it is not being used (yet)

AybukeKY (approved):

PR #126, opened 2024-07-23
removed dataset type enums for validation

sherwoodf:
These enums were created to deal with different types at the end of the submitted_in_dataset_uuid paths, but we think this could be better handled by using two subclasses (thought this might introduce the same issue with the link from derived image -source_image-> file representation). In any case, this should probably be looked into when desiging the UUID validation & have more thinking about it (perhaps a design doc). Removing for now, and cut this ticket https://app.clickup.com/t/86956e5g9 to track the follow-up work.


liviuba (approved):
LGTM!

PR #127, opened 2024-07-24
disallowed extra fields in models, and updated ingest and export code to handle these

sherwoodf:
See: https://app.clickup.com/t/86956q84r

Also updated website export code to use input that are at least as complex as the BIA data models (i.e. with the various api fields), since it will make using the clients easier.

### DIFF
--- a/bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
+++ b/bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
@@ -151,3 +151,9 @@ def persist(object_list: List, object_path: str, sumbission_accno: str):
         output_path = output_dir / f"{object.uuid}.json"
         output_path.write_text(object.model_dump_json(indent=2))
         logger.info(f"Written {output_path}")
+
+
+def filter_model_dictionary(dictionary: dict, target_model: Type[BaseModel]):
+    accepted_fields = target_model.model_fields.keys()
+    result_dict = {key: dictionary[key] for key in accepted_fields}
+    return result_dict
--- a/bia-ingest-shared-models/test/utils.py
+++ b/bia-ingest-shared-models/test/utils.py
@@ -6,8 +6,10 @@ This module attempts to create models starting from the outer nodes (leaves) of 
 
 from typing import Dict, List
 from bia_shared_datamodels import bia_data_model, semantic_models
-from bia_ingest_sm.conversion.utils import dict_to_uuid
-
+from bia_ingest_sm.conversion.utils import (
+    dict_to_uuid,
+    filter_model_dictionary
+)
 
 def get_test_annotation_method() -> List[bia_data_model.AnnotationMethod]:
     # For UUID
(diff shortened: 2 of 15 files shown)


bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
```diff
@@ -150,4 +150,10 @@ def persist(object_list: List, object_path: str, sumbission_accno: str):
     for object in object_list:
         output_path = output_dir / f"{object.uuid}.json"
         output_path.write_text(object.model_dump_json(indent=2))
-        logger.info(f"Written {output_path}")
\ No newline at end of file
+        logger.info(f"Written {output_path}")
+
+
+def filter_model_dictionary(dictionary: dict, target_model: Type[BaseModel]):
```
sherwoodf, line 156: I wonder whether this has utility beyond just the export code, and would be better off as a utils in the shared_models package. Thoughts?

liviuba, line 156: Nice workaround for the filter!

liviuba (commented):
filter_model_dictionary looks good, I wouldn't know about how it fits with the rest of the code though

bia-ingest-shared-models/test/utils.py
```diff
@@ -34,47 +36,12 @@ def get_test_annotation_method() -> List[bia_data_model.AnnotationMethod]:
...
     ]
 
-    protocol = []
-    for protocol_dict in protocol_info:
-        protocol_dict["uuid"] = dict_to_uuid(protocol_dict, attributes_to_consider)
-        protocol.append(bia_data_model.AnnotationMethod.model_validate(protocol_dict))
-    return protocol
-
-
-def get_test_specimen_growth_protocol() -> List[bia_data_model.ImageAcquisition]:
```
sherwoodf, line 44: This whole method was just a duplicate, so i removed it.

bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
kbab: 1. Are you referring to `persist` or `filter_model_dictionary`
2. Do you mean 'import code' in comment above? It seems both `import` and `export` use `persist` - So having it in a common place sounds good. However, I assume `export` is only reading  data models so will not need the filter functionality (but will need the persisting).

sherwoodf: 1. filter_model_dictionary
2. i meant import. But the export could use it with slightly different data models (e.g. the study model for export is the study model + dataset model + biosample, protocols, Image Acquisitions etc.) so having a filter model dictionary method might be helpful.

3. For persist - While both might want to write out to files, it feels like the directory/file structure might be very different for the export and ingest code, so i don't know how reusable the methods would be. Additionally, the functions related to filtering models makes sense to store in the model package since it's related and is going to be imported by both, but it's less obvious where we should put the persisting.

PR #129, opened 2024-07-26
updated example study output

sherwoodf:
Updated example output from study conversion.

Does not yet contain versions etc from Liviu's open PR.

### DIFF
--- a/bia-ingest-shared-models/test/data/S-BIAD606/output.json
+++ b/bia-ingest-shared-models/test/data/S-BIAD606/output.json
@@ -30,37 +30,15 @@
             "development",
             "segmentation"
         ],
-        "funding_statement": "Deutsche Forschungsgemeinschaft grant BR5039/3-1\nAssemble Plus grant as part of the European Union’s Horizon 2020 research and innovation programme under grant agreement no. 730984",
-        "experimental_imaging_component": [
-            "UUID-experimental-imaging-dataset-1"
-        ],
-        "annotation_component": []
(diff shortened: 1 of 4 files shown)


bia-ingest-shared-models/test/data/S-BIAD606/output.json
```diff
@@ -95,27 +73,23 @@
...
+        ],
+        "sample_of": [
+            "UUID-biosample-1"
+        ],
+        "growth_protocol": []
+    },
     {
         "uuid": "UUID-image-representation-1",
+        "representation_of": "UUID-experimentally-captured-image-1",
         "original_file_reference": [
```
sherwoodf, line 93: I'm leaning towards linking from Image Representation -> file reference, even if this will potentially be 1: A few.

I suspect File Reference -> Image Representation would be 1:1 most of the time judging by current code (only 1 image rep. gets linked back to a file reference - i'm assuming this is expected)

However, this feels better from the point of view of containing create/update operations. If you create a new image rep, you can link back to the original file references. This seems preferable compared to having to go the original file references and update them to point to the new image rep. Thoughts?

bia-ingest-shared-models/test/data/S-BIAD634/output.json
```diff
@@ -170,47 +170,15 @@
...
-            "UUID-biosample-6"
-        ],
-        "specimen_growth_protocol": [],
         "analysis_method": [],
         "correlation_method": [],
-        "file_reference_count": 1,
-        "image_count": 1,
         "example_image_uri": [],
         "attribute": {
             "associatations": [
```
AybukeKY, line 184: Typo: associatations written instead of associations!

sherwoodf:
Closing this - will updated propperly once models have gone through more review

PR #130, opened 2024-07-30
CI Testing multiple projects (not clean)

ctr26:
@sherwoodf Is this what you were thinking?

Added a matrix to the current CI so that it tests multiple folders. Negative here is that someone will need to edit the GitHub ci to add anymore subproject to testing

### DIFF
--- a/.github/workflows/core.yaml
+++ b/.github/workflows/core.yaml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     defaults:
       run:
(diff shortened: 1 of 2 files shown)


sherwoodf:
ooo, kind of, i was trying to create one workflow per package, but i'm no expert on setting up these workflows (currently trying to get act working to test what i've made...) If this works and runs all the tests it sounds good to me.

sherwoodf (approved):
LGTM!

PR #132, opened 2024-07-30
Dockercompose to unify all the docker components

ctr26:
@liviuba 

I've made a start on moving the docker-compose file to the root dir. This should improve readability, expandability and fix the context issue. 

Also solves the issue of what files are ignored by docker ignore

This context issue is that the api project can't see files from the other projects unless the context is set correctly. 

Hope this helps,


clients/Readme.md
```diff
@@ -4,28 +4,8 @@
...
-
-```sh
-docker run -it --rm \
-    -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro --user $(id -u) \
-    -v ${PWD}:/mnt/share \
-    -w /mnt/share \
-    openapitools/openapi-generator-cli:v7.0.0 \
-    config-help -g python
-```
+docker-compose run generate-client
```
liviuba, line 7: ty! Do you have a strong preference for docker-compose? IIRC there was an issue a long time ago where the api wouldn't start I think when starting with docker-compose, but it did with `docker compose`, turns out Docker deprecated docker-compose (IIRC) and the docker-compose file version was unsupported

I'd need to double-check details but basically the conclusion was that I'm always using `docker compose` now

ctr26: I did not know their was a difference tbh, I thought one was a symlink to the other

liviuba: just checked on my local and it a binary (but my local might be inconsistent / have an old install) `-rwxr-xr-x 1 root root 420 Nov 23  2019 /usr/bin/docker-compose`

I think docker-compose is there just for compatibility or something

ctr26:
@liviuba, I rebased on api_shared_datamodels when I maybe shouldn't have as I have commits from other branches in here too.  Making this one a draft as a result

ctr26:
@liviuba @sherwoodf 

This PR depends on

https://github.com/BioImage-Archive/bia-integrator/pull/134

now

ctr26:
Also throwing errors on docker compose tests which I think is expected

clients/Readme.md
ctr26: Thanks!

ctr26:
@liviuba 

`api/tests/test_dummy.py F  `

Otherwise I'm happy for this to be merged if you are.

ctr26:
Necessity is the mother of invention.

I removed the extra commits' code that snuck in by using

```
git rebase -i origin/main
then I changed all the commits that I didn't write to edit
git reset HEAD~1
git commit --amend
gitr rebase --continue
```

ctr26:
Fun one, 

Just now I tried to rebase my local api_shared_datamodels on this branch and there was a merge conflict because I tried the older docker compose instead of `git mv`

I then went back into the git history and modified that from a delete and create to a  `git mv` and now there's no rebase merge errors on 

`git rebase -X ours origin/dockercompose`

api/Dockerfile
```diff
@@ -3,27 +3,25 @@ FROM python:3.10.0
...
+RUN poetry lock
 RUN poetry install
 
 # Everything up to here should be reused most times
 
-# add the actual project, which is what is often changed in between two different container builds
-ADD ./ /integrator-api
 
-CMD ["poetry", "run", "uvicorn", "--workers", "4", "--port", "8080", "--log-config", "./src/log_config.yml", "--host", "0.0.0.0", "src.app:app"]
+CMD ["poetry", "run", "uvicorn", "--workers", "4", "--port", "8080", "--log-config", ".api/src/log_config.yml", "--host", "0.0.0.0", "api.app:app"]
```
liviuba, line 27: Typo for `./api...` here?

liviuba (approved):
LGTM!

Not sure about the first commit on this branch (72585733bf150e38ab533721d702062d7552b8aa) having the same message as an older squashed PR, but no conflicts so assuming nothing is overwritten that it shouldn't

api/Dockerfile
ctr26: I'm too good at fixing files in the git history so now I'll never look like I've made a mistake ever again lol

PR #134, opened 2024-07-31
Api shared datamodels

liviuba:
To test:
* `docker compose up -d --build`
* go to http://localhost:8080/docs and create/get objects

Actual changes:
* the one test that creates/gets a study (`tests/test_minimal.py`)
* `api/private`, `api/public`

### DIFF
--- a/api/api/private.py
+++ b/api/api/private.py
@@ -14,10 +14,19 @@ router = APIRouter(
 )
 models_private = [
     shared_data_models.Study,
-    shared_data_models.ImageAnnotationDataset,
-    shared_data_models.ExperimentalImagingDataset,
-    shared_data_models.AnnotationFileReference,
     shared_data_models.FileReference,
+    shared_data_models.ImageRepresentation,
+    shared_data_models.ExperimentalImagingDataset,
+    shared_data_models.Specimen,
+    shared_data_models.ExperimentallyCapturedImage,
+    shared_data_models.ImageAcquisition,
+    shared_data_models.SpecimenImagingPrepartionProtocol,
+    shared_data_models.SpecimenGrowthProtocol,
+    shared_data_models.BioSample,
+    shared_data_models.ImageAnnotationDataset,
+    shared_data_models.AnnotationFileReference,
+    shared_data_models.DerivedImage,
+    shared_data_models.AnnotationMethod,
 ]
 
 
--- a/api/api/private.py
+++ b/api/api/private.py
@@ -5,7 +5,7 @@ from pydantic.alias_generators import to_snake
 import bia_shared_datamodels.bia_data_model as shared_data_models
 from .models.repository import Repository
 from . import constants
-from fastapi import APIRouter, Depends
+from fastapi import APIRouter, Depends, status
 
 router = APIRouter(
     prefix="/private",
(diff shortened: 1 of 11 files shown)


ctr26:
@liviuba I've got most of the docker compose changes here in the docker compose branch and I've rebased on this branch to merge them

liviuba:
> @liviuba I've got most of the docker compose changes here in the docker compose branch and I've rebased on this branch to merge them

Thanks, will look at it! Do you want to merge into this before we merge this into main to avoid a re-rebase if we squash&merge this? If you want to merge into main, then I can fastforward main to this branch after the PR is done to merge, and then the docker-compose one shouldn't need another rebase

ctr26:
Happy to it's just got some of your API code in it too now, so it won't be completely clean, but it's all additive so it won't cause obvious problems

api/api/public.py
```diff
@@ -0,0 +1,47 @@
+from fastapi import APIRouter, Depends
+from pydantic.alias_generators import to_snake
+
+# ?
+import bia_shared_datamodels.bia_data_model as shared_data_models
+from .models.repository import Repository
+
+
+router = APIRouter()
```
sherwoodf, line 9: should the be using the public tag, like the private one is?

api/api/private.py
```diff
@@ -0,0 +1,48 @@
+from fastapi import APIRouter
+from pydantic.alias_generators import to_snake
+
+# ?
```
sherwoodf, line 4: ?

api/api/public.py
```diff
@@ -0,0 +1,47 @@
...
+    # https://eev.ee/blog/2011/04/24/gotcha-python-scoping-closures/
+
+    # @TODO: nicer wrapper?
+    async def get_item(uuid: shared_data_models.UUID, db: Repository = Depends()) -> t:
+        return await db.get_doc(uuid, t)
+
+    return get_item
+
+
+for t in models_public:
```
sherwoodf, line 39: am i right in thinking that this gets run when app.py imports the public and private endpoints? Just wondering whether it would be better python practice to have methods that get explicitly called in app.py...?

sherwoodf (commented):
Generally looks good, but left a couple of comments. Checked it out and it builds locally with the docker command & was able to post & get a study (only got some errors due to the problems with model).

Was able to run the test 'manually' successfully but had issues with pydantic trying to find them - that's going to just be user error though.

api/api/public.py
liviuba: Yep, I think I missed it, fixed

api/api/private.py
liviuba: Haha, must have added it as a note because I wasn't sure if importing directly would work out, deleted now

api/api/public.py
liviuba: Agree it would be nice to not run code in global scope at import, was just handwaving the problem until I get an example of a non-generated handler too, so I can see how I'd like to do it but still keep router global (to avoid nesting custom handlers in the factory), and make "the correct way" to use a router easy to tell (e.g. no import first and then a function to attach the generation functionality)

Added a fix, but I'm not sure we'll keep it like this once we have more functionality

sherwoodf (approved):
LGTM

PR #128, opened 2024-07-25
Shared models mongofy

liviuba:
For https://app.clickup.com/t/86956qv26
Not tested, api still wip

PR #135, opened 2024-07-31
fix tests after model update with version and model version, and extend forbidding extra fields to nested nodes

sherwoodf:
Fixes tests broken by: https://github.com/BioImage-Archive/bia-integrator/pull/128#pullrequestreview-2208119050

Also moves the configuration that forbids extra fields into the semantic models file to be inherited instead of BaseModel. This stops extra fields from being passed into models that govern 'nested'/'embedded' json objects inside other json objects

Also updates the filter model dictionary function to work with/expect optional fields.

### DIFF
--- a/bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
+++ b/bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
@@ -155,5 +155,5 @@ def persist(object_list: List, object_path: str, sumbission_accno: str):
 
 def filter_model_dictionary(dictionary: dict, target_model: Type[BaseModel]):
     accepted_fields = target_model.model_fields.keys()
-    result_dict = {key: dictionary[key] for key in accepted_fields}
+    result_dict = {key: dictionary[key]  for key in accepted_fields if key in dictionary}
     return result_dict
--- a/bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
@@ -25,9 +25,6 @@ class DocumentMixin(BaseModel):
         Optional because for some usecases (e.g. api) we want to accept objects without it because we have the info we need to set it."""
     )
 
-    # Throw error if you try to validate/create model from a dictionary with keys that aren't a field in the model
-    model_config = ConfigDict(extra="forbid")
-
     def __init__(self, *args, **data):
         model_version_spec = self.model_config.get("model_version_latest")
         if model_version_spec is None:
@@ -41,6 +38,7 @@ class DocumentMixin(BaseModel):
         )
         model_metadata_existing = data.get("model", None)
         if model_metadata_existing:
+            model_metadata_existing = ModelMetadata(**model_metadata_existing)
             if model_metadata_existing != model_metadata_expected:
                 raise exceptions.UnexpectedDocumentType(
                     f"Document {str(data.get('uuid'))} has model metadata {model_metadata_existing}, expected : {model_metadata_expected}"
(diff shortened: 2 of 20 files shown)


bia-ingest-shared-models/bia_ingest_sm/conversion/utils.py
```diff
@@ -155,5 +155,5 @@ def persist(object_list: List, object_path: str, sumbission_accno: str):
 
 def filter_model_dictionary(dictionary: dict, target_model: Type[BaseModel]):
     accepted_fields = target_model.model_fields.keys()
-    result_dict = {key: dictionary[key] for key in accepted_fields}
+    result_dict = {key: dictionary[key]  for key in accepted_fields if key in dictionary}
```
sherwoodf, line 158: Fix to cope with optional/missing fields in the original dictionary

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
```diff
@@ -2,17 +2,22 @@
...
-from typing import List, Optional, Union
+from typing import List, Optional
 
-from pydantic import BaseModel, Field, EmailStr, AnyUrl
+from pydantic import BaseModel, Field, EmailStr, AnyUrl, ConfigDict
+
+
+class ConfiguredBaseModel(BaseModel):
+    # Throw error if you try to validate/create model from a dictionary with keys that aren't a field in the model
+    model_config = ConfigDict(extra="forbid")
```
sherwoodf, line 12: Moved to the semantic models to make sure there isn't an edge case where e.g. a study can't have extra fields, but a contributor in a study can

bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
```diff
@@ -41,6 +38,7 @@ def __init__(self, *args, **data):
         )
         model_metadata_existing = data.get("model", None)
         if model_metadata_existing:
+            model_metadata_existing = ModelMetadata(**model_metadata_existing)
```
sherwoodf, line 41: Fix to the error checking - as it was currently trying to compare a dictionary with a model. This was in the original API models but was missed when copied over to here.

bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
```diff
@@ -25,9 +25,6 @@ class DocumentMixin(BaseModel):
         Optional because for some usecases (e.g. api) we want to accept objects without it because we have the info we need to set it."""
```
kbab, line 25: Is this optional "because we have the info..." or "because we do _not_ have the info ..."

kbab: Ahh - I understand now after looking at the code in __init__

kbab (approved):
LGTM

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
liviuba: btw nice find, we've literally had this bug in the api for nested objects since forever

liviuba (approved):
LGTM!

bia-shared-datamodels/test/utils.py
```diff
@@ -196,6 +234,11 @@ def get_template_image_acquisition() -> bia_data_model.ImageAcquisition:
             "fbbi_id": [
                 "Test FBBI ID",
             ],
+            "version": 1,
+            "model": {
+                "type_name": "ImageAcquisition", 
+                "version": 1
+            }
```
liviuba, line 241: Actually, there's an error in the CI https://github.com/BioImage-Archive/bia-integrator/actions/runs/10197920825/job/28211773906?pr=135 just saw it, but it's from a few lines below this

PR #136, opened 2024-08-01
example new test format

sherwoodf:
Would like to get some feedback on this test / util structure before i go change all the util methods, as that will take a while.

1. Created a COMPLETE/MINIMAL enum in the utils
2. Instead of the utils methods returning a 'study' or other object, they return a dictionary. They also accept the enum value to modify the dictionary that gets returned.
3. The tests use the dictionary method to get UUIDs for linked objects (i.e. through get_study_dict()["uuid"])- this keeps the dependencies between the different objects obvious, but should reduce blanket test failing when models are changed, making debugging easier.

Also leaving a comment about minimal testing missing out some objects in the code - it's clearer to see in context.

Model changes: these are to a) to make definitions that are supposed to behave the same, look the same, and b) fix issues where models were partially updated (e.g. just the annotation, but not the default) causing the <instance of class>.model_dump() to not create a dicitionary that would pass the class validation

_____

After discussion: we're happy with the new format, and i have updated all tests to follow that format, and so updated all the util methods to follow the conventions.

Additional small fixes to the model.

Also added testing for 'minimum list length' of fields as part of the minimal model tests.

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
@@ -32,14 +32,14 @@ class Study(ConfiguredBaseModel):
     )
     release_date: date = Field(description="""Date of first publication""")
     description: str = Field(
-        None, description="""Brief description of the study."""
+        description="""Brief description of the study."""
     )
     keyword: Optional[List[str]] = Field(
         default_factory=list,
         description="""Keywords or tags used to describe the subject or context of the study.""",
     )
     acknowledgement: Optional[str] = Field(
-        default_factory=list,
+        None,
         description="""Any person or group that should be acknowledged outside of the authors/main contributors to the study.""",
     )
 
@@ -55,7 +55,7 @@ class Study(ConfiguredBaseModel):
         default_factory=list, description="""The grants that funded the study."""
     )
     funding_statement: Optional[str] = Field(
-        default_factory=list, description="""Description of how the study was funded."""
+        None, description="""Description of how the study was funded."""
     )
 
     # TODO: In order to maintian consistency these will be endpoints that run a query in the DB, rather than a stored field.
@@ -164,7 +164,7 @@ class OrganisationMixin(ConfiguredBaseModel):
         None, description="""Comma separated lines of the address."""
     )
     website: Optional[AnyUrl] = Field(
-        default=None,
+        None,
         description="""The website page with information about the Organisation.""",
     )
 
@@ -178,14 +178,13 @@ class Contributor(PersonMixin, OrganisationMixin):
         description="""Name as it should be displayed on the BioImage Archive."""
     )
     affiliation: List[Affiliation] = Field(
-        default_factory=list,
         description="""The organisation(s) a contributor is afiliated with.""",
     )
     contact_email: Optional[EmailStr] = Field(
-        default=None, description="""An email address to contact the Contributor."""
+        None, description="""An email address to contact the Contributor."""
     )
     role: Optional[str] = Field(
-        default=None, description="""The role of the contributor."""
+        None, description="""The role of the contributor."""
     )
 
 
@@ -329,7 +328,7 @@ class ImageRepresentation(ConfiguredBaseModel):
         description="""Size of temporal dimension of the data array of the image.""",
     )
     image_viewer_setting: Optional[List[RenderedView]] = Field(
-        None,
+        default_factory=list,
         description="""Settings of a particular view of an image, such as a specific timestamp of a timeseries, or camera placement in a 3D model.""",
     )
     # TODO: representation_of information is stored in representation_of_uuid defined in bia_data_model. 
@@ -356,7 +355,7 @@ class RenderedView(ConfiguredBaseModel):
         None, description="""A t-value for the timestamp of the image view"""
     )
     channel_information: Optional[List[Channel]] = Field(
-        None,
+        default_factory=list,
         description="""Information about the channels involved in displaying this view of the image.""",
     )
 
@@ -399,9 +398,11 @@ class ExperimentalImagingDataset(DatasetMixin):
     #     description="""Processes involved in the growth of the samples that were then imaged."""
     # )
     analysis_method: Optional[list[ImageAnalysisMethod]] = Field(
+        default_factory=list,
         description="""Data analysis processes performed on the images."""
     )
     correlation_method: Optional[list[ImageCorrelationMethod]] = Field(
+        default_factory=list,
         description="""Processes performed to correlate image data."""
     )
     example_image_uri: list[str] = Field(
@@ -441,6 +442,7 @@ class ImageAcquisition(ProtocolMixin):
         description="""Names, types, or description of how the instruments used to create the image."""
     )
     fbbi_id: Optional[List[str]] = Field(
+        default_factory=list,
         description="""Biological Imaging Methods Ontology id indicating the kind of imaging that was perfomed."""
     )
     imaging_method_name: Optional[str] = Field(
@@ -453,7 +455,10 @@ class SpecimenImagingPrepartionProtocol(ProtocolMixin):
     The process to prepare biological entity for imaging.
     """
 
-    signal_channel_information: Optional[List[SignalChannelInformation]]
+    signal_channel_information: Optional[List[SignalChannelInformation]] = Field(
+        default_factory=list,
+        description="""Information about how channels in the image relate to image signal generation."""
+    )
 
 
 class SignalChannelInformation(ConfiguredBaseModel):
@@ -512,12 +517,15 @@ class BioSample(ConfiguredBaseModel):
         description="""A short description of the biological entity."""
     )
     experimental_variable_description: Optional[List[str]] = Field(
+        default_factory=list,
         description="""What is intentionally varied (e.g. time) between multiple entries in this study component"""
     )
     extrinsic_variable_description: Optional[List[str]] = Field(
+        default_factory=list,
         description="External treatment (e.g. reagent)."
     )
     intrinsic_variable_description: Optional[List[str]] = Field(
+        default_factory=list,
         description="Intrinsic (e.g. genetic) alteration."
     )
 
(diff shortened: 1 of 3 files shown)


bia-shared-datamodels/test/utils.py
```diff
@@ -359,57 +342,82 @@ def get_template_image_representation() -> bia_data_model.ImageRepresentation:
...
-def get_template_affiliation() -> semantic_models.Affiliation:
-    affiliation = semantic_models.Affiliation.model_validate(
-        {
+def get_affiliation_dict(
+    completeness=Completeness.COMPLETE,
+) -> dict:
+
+    if completeness == Completeness.MINIMAL:
+        affiliation = {
+            "display_name": "Template Affiliation Organisation",
```
sherwoodf, line 351: Minimal affiliation doesn't get tested here, because the contributor object doesn't require one to be present. I think it might be better to test every object individually, even if that tests some twice (because they are nested), to make sure they all pass.

bia-shared-datamodels/test/utils.py
```diff
@@ -359,57 +342,82 @@ def get_template_image_representation() -> bia_data_model.ImageRepresentation:
...
             "accession_id": "S-BIADTEST",
             "licence": semantic_models.LicenceType.CC0,
+            "author": [get_contributor_dict(Completeness.MINIMAL)],
             "attribute": {},
-            "related_publication": [],
-            "author": [
-                contributor.model_dump(),
-            ],
             "title": "Test publication",
             "release_date": "2024-06-23",
```
sherwoodf, line 395: I forgot that due to datatypes,

"release_date": "2024-06-23"

would get turned into:

"release_date": datetime.date(2024, 6, 23).

Testing the string version in the minimal works nicely with the tests i'm proposing. Is it confusing to have both?

bia-shared-datamodels/test/test_shared_models.py
```diff
@@ -75,6 +70,38 @@ def test_create_models(expected_model_type, model_creation_func):
     assert type(expected_model) is expected_model_type
 
 
+def test_create_study():
```
sherwoodf, line 73: If we're happy with this test, i'll expand it to cover all the classes / replace the test above.

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
```diff
@@ -453,7 +455,10 @@ class SpecimenImagingPrepartionProtocol(ProtocolMixin):
     The process to prepare biological entity for imaging.
     """
 
-    signal_channel_information: Optional[List[SignalChannelInformation]]
+    signal_channel_information: Optional[List[SignalChannelInformation]] = Field(
+        default_factory=list,
```
ctr26, line 459: My reading of this is that the field is optional but if there is a List then it should contain SignalChannelInformation

Meaning it should be 

Optional[List[Optional[SignalChannelInformation]]] ?

for 

default_factory=list

?

ctr26: Or even

List[Optional[SignalChannelInformation]]

Considering there's a default factory?

sherwoodf: Oh right - this is very confusing & definitely not what LinkML generates, but there are enough issues with what linkml generates that i don't know to trust this.

I think it's pydantic that doesn't do what you would expect it to be. Some examples:
```
affiliation: List[Affiliation] = Field(
        description="""The organisation(s) a contributor is afiliated with.""",
    )
```
Allows [] to be set as a value. Whereas:
```
    author: List[semantic_models.Contributor] = Field(min_length=1)
```
Does actually require at least one author to be set.

Also!! The field actually isn't optional even if it's Optional[List] - what makes it option is having a default or default_factory.

ctr26: Thanks!

sherwoodf: I don't know whether there's enough leeway that we should write the annotation to be exactly what we expect. Is this pydantic or python being weird though? I've seen this question a lot online: https://stackoverflow.com/questions/60191270/python-pydantic-how-to-have-an-optional-field-but-if-present-required-to-con

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
```diff
@@ -399,9 +398,11 @@ class ExperimentalImagingDataset(DatasetMixin):
     #     description="""Processes involved in the growth of the samples that were then imaged."""
     # )
     analysis_method: Optional[list[ImageAnalysisMethod]] = Field(
```
liviuba, line 400: it's not part of the actual pr, but could this be List? Can't think of an example now but I'm worried small mismatches like this will cause issues later

bia-shared-datamodels/test/test_shared_models.py
liviuba: looks great, just 2 questions
1. Is this intended to be depended on by other modules? get_study_dict (&similar) would be very useful in api tests
2. doesn't affect LOC, but IMO it would be nice if we had a convention like get_study_dict returns a dictionary and get_study returns the line below

liviuba (commented):
Looks good, if I could reuse this in the api it would be amazing

bia-shared-datamodels/test/utils.py
kbab: It seems OK over here - at least we are testing the form in which we expect it from the biostudies submission (the string version)

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
kbab: I am not sure what the correct answer is - however, I think consistency on our side will be helpful. I.e have our own convention and use it consistently.

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
sherwoodf: That should definitely be List

bia-shared-datamodels/test/test_shared_models.py
sherwoodf: > 1. Is this intended to be depended on by other modules? get_study_dict (&similar) would be very useful in api tests

I had thought it could be useful for other test suites, but i've not thought about the implications to say this will definitely work for it. Feel free to try to re-use them for tests if that helps!

> get_study_dict returns a dictionary and get_study returns the line below

Makes sense. But also +1  to the LOC question XD

kbab (commented):
IMO it is useful to have the COMPLETE/MINIMAL tests. The use of the dictionary methods also adds flexibility. I left some other comments at specific lines, but overall LGTM.

bia-shared-datamodels/test/test_shared_models.py
liviuba: oops my bad, it's lines of code, as in it wouldn't be an actual utility function that reduces the number of lines in downstream code, it just felt nice to have "function that returns thing" be both "dict thing" and "object thing" next to each other

kbab: @liviuba out of curiosity, would a refactor be required to reuse in API or would you simply import the utils module from this package?

liviuba: I think I can just import, this is the upside of having just 1 test, makes it easy to change! :face_with_head_bandage: 

 I'll try it out, it would be cool if we can make models easier to change independently and most times in the api tests I just need anything that's just a valid object of that type

sherwoodf:
> IMO it is useful to have the COMPLETE/MINIMAL tests. The use of the dictionary methods also adds flexibility. I left some other comments at specific lines, but overall LGTM.

Nice! Thanks for reviewing. I will go through and update all the tests to follow this pattern.

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
sherwoodf: found a few of these and fixed them! Good spot

bia-shared-datamodels/test/test_shared_models.py
```diff
@@ -1,97 +1,94 @@
...
+    # Check that the model is created
+    assert type(complete_model) is expected_model_type
+    # Check that the dictionary is indeed "Complete" - no optional fields missed
+    assert complete_model.model_dump() == complete_dict
+    # Check that there are no inconsistencies in the model definition
+    assert (
+        type(expected_model_type(**complete_model.model_dump())) is expected_model_type
+    )
+
+    mimimal_dict = dict_creation_func(utils.Completeness.MINIMAL)
```
liviuba, line 75: Can this be split into a separate test? Something like test_create_complete and test_create_minimal?

bia-shared-datamodels/test/utils.py
```diff
@@ -6,100 +6,133 @@
...
+        specimen_imaging_preparation_protocol = {
+            "uuid": uuid4(),
+            "title_id": "Test specimen preparation protocol",
+            "protocol_description": "Test description",
+            "version": 1,
+        }
+    elif completeness == Completeness.COMPLETE:
+        specimen_imaging_preparation_protocol = {
+            "uuid": uuid4(),
+            "title_id": "Test specimen preparation protocol",
```
liviuba, line 88: just because these functions will tend to get large and with lots of fields that are hard to track, it might be nice if we did something like

```py
def get_something_dict(...):
    thing_dict = {
        ...minimal example
    }
    if COMPLETE:
        thing_dict |= {
            ...optional fields
        }
```

this way if we make changes to the models, it's easier to tell if something is missing or not (between the two dicts I mean)

liviuba (approved):
LGTM, left some suggestions to improve maintainability

bia-shared-datamodels/test/test_shared_models.py
sherwoodf: It will duplicate all the parameters, but should be fine.

bia-shared-datamodels/test/utils.py
sherwoodf: .... we could, but we do override some fields in the complete ones e.g. using string date vs datetime date in the study or adding values to list fields. So it won't be LOC shorter by X where X is the number of optional fields...

The benefit of having the full dicts is that we can easily copy and paste them as a starting point for other tests.

bia-shared-datamodels/test/test_shared_models.py
liviuba: oh, didn't think of that, fair, but can the function be renamed to something more suggestive and left like that then? You're right that it's not worth the messyness, it just needs to be clearer IMO. Maybe a block comment-type thing for either option, or splitting the actual tests in individual functions (one for min and one for complete) for clarity?

PR #137, opened 2024-08-02
added export of associated objects from dataset

sherwoodf:
See bia-export/test/output_data/bia_export.json for example of what now gets exported.

Export uses associations - since until we have the 'reverse' endpoints, there isn't a direct connection between Datasets and Biosamples, Image Acquistions, etc...


bia-export/bia_export/website_conversion.py
```diff
@@ -7,44 +7,101 @@
...
+    """
+    Returns model of object from file to be equivalent to using BIA API Client
+    """
+    with open(file_path, "r") as object_file:
+        object_dict = json.load(object_file)
+    
+    return object_type(**object_dict)
+
+
+def real_all_json(directory_path: Path, object_type: Type[BaseModel]) -> List[BaseModel]:
```
AybukeKY, line 27: Did you mean read_all_json ? Or reel_all_json??

sherwoodf:
Also, test case json results in this being displayed on the website!

image

bia-export/bia_export/website_conversion.py
sherwoodf: haha read_all_json

bia-export/bia_export/website_models.py
```diff
@@ -8,4 +8,16 @@ class Study(bia_data_model.Study):
...
+    acquisition_process: list[bia_data_model.ImageAcquisition] = Field(
+        description="""Processes involved in the creation of the images and files in this dataset."""
+    )
+    specimen_imaging_preparation_protocol: list[bia_data_model.SpecimenImagingPrepartionProtocol] = Field(
+        description="""Processes involved in the preprapartion of the samples for imaged."""
+    )
+    biological_entity: list[bia_data_model.BioSample] = Field(
+        description="""The biological entity or entities that were imaged."""
+    )
+    specimen_growth_protocol: Optional[list[bia_data_model.SpecimenImagingPrepartionProtocol]] = Field(
```
AybukeKY, line 20: specimen_growth_protocol refers to SpecimenImagingPreparationProtocol instead of the SpecimenGrowthProtocol which I can see in the bia shared data models. Also in the shared data models there's Specimen but that doesn't seem to be updated as it doesn't refer to the correct protocols and it's sort of commented out in the semantic models.

bia-export/bia_export/website_conversion.py
```diff
@@ -7,44 +7,101 @@
...
 
-            with open(eid_path, "r") as eid_file:
-                eid_dict = json.load(eid_file)
+            biosample_directory = root_directory.joinpath(f'biosamples/{accession_id}/*.json')
+            sipp_directory = root_directory.joinpath(f'specimen_imaging_preparation_protocols/{accession_id}/*.json')
+            sgp_directory = root_directory.joinpath(f'specimen_growth_protocols/{accession_id}/*.json')
+            ia_directory = root_directory.joinpath(f'image_acquisitions/{accession_id}/*.json')
+
+            biosample_linked_to_dataset = find_associated_objects(association_by_type["biosample"], biosample_directory, bia_data_model.BioSample)
+            sipp_linked_to_dataset = find_associated_objects(association_by_type["specimen"], sipp_directory, bia_data_model.SpecimenImagingPrepartionProtocol)
```
AybukeKY, line 94: Do we want to put the logic of taking out specimen preparation and growth protocols here or do we want to take everything via specimen (whatever specimen in the models include)?

bia-export/bia_export/website_models.py
sherwoodf: I don't know what you mean about specimen, but will fix this to SpecimenGrowthProtocol

bia-export/bia_export/website_conversion.py
sherwoodf: What do you mean by 'taking out'?

We can't use Specimen in the 'new data model' sense, because that's connected to images, and i'm trying to avoid searching over all images in a dataset.

AybukeKY: Ah ok. I thought Specimen was like a container that links to growth protocol and preparation protocol but I think that part is either not set yet or is now completely different with the whole orange arrows discussion and/or tied to images which then requires searching over images as you've said.

AybukeKY: Now looking at the other PR (129); helps clarify things actually.

AybukeKY (approved):

PR #140, opened 2024-08-06
added default_open field to details section of datasets

sherwoodf:
In the website these are all using html 'details' sections. We want these to be open if it's the first time a e.g. biosample is being displayed, but be closed if a different section re-references the same biosample.

I think it is easier to store this field in the json rather than try to get Astro's rendering logic work out which sections it has rendered before. This PR adds 'default_open' field to BioSample etc. models for use in the website (named that way because True/False maps directly to the value needed for the html)

This PR doesn't cover Image analysis and correlation sections that are embedded objects in the Dataset json - i will revisit that separately as it's a bit more complex. Also refactored some code into a function because i was having to do the same logic for each of the objects.

Part of: https://app.clickup.com/t/86956nvk4

Astro website PR using this field: https://github.com/BioImage-Archive/BIA-astro/pull/13

### DIFF
--- a/.github/workflows/core.yaml
+++ b/.github/workflows/core.yaml
@@ -39,4 +39,4 @@ jobs:
       - name: Check to see if poetry can build
         run: poetry build
       - name: Run pytest
-        run: poetry run pytest
+        run: poetry run pytest -s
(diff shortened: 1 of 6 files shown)


AybukeKY (approved):
Approving with a note that the input data form/logic may change but that shouldn't affect the previously_set logic.

PR #141, opened 2024-08-06
Fixing more docker related stuff

ctr26:
- Old changes broke the deployment for the main api image build.
- Issue was that the CI was quite specific, modified it to be more general
- Also make the building caching faster

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -19,7 +19,7 @@ ADD bia-shared-datamodels bia-shared-datamodels
 WORKDIR /bia-integrator/api
 
 # This will now cache unless theres changes to the lock file, or the pyproject.toml, or bia-shared-datamodels
-RUN poetry install
+RUN poetry install --no-root
 
 # Add the rest of the files
 # The previous layers will cache fully and only this layer will be rebuilt if there are changes to the files.
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -12,6 +12,8 @@ COPY . .
 
 WORKDIR /bia-integrator/api
 
+# Poetry lock command likely unnecessary as we are copying the whole project
+# For local development we only need to solve the lock once
 RUN poetry install
 
 CMD ["poetry", "run", "uvicorn", "--workers", "4", "--port", "8080", "--log-config", ".api/src/log_config.yml", "--host", "0.0.0.0", "api.app:app"]
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -6,20 +6,12 @@ EXPOSE 8080
 RUN curl -sSL https://install.python-poetry.org | python3 -
 ENV PATH=/root/.local/bin:$PATH
 
-# add files that poetry needs first so layers up to (including) pulling dependencies get reused
 WORKDIR /bia-integrator
 
-#   only add poetry.lock if it exists (building on local)
-ADD ./api/poetry.lock* api/poetry.lock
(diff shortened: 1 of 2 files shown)


ctr26:
https://github.com/ctr26/bia-integrator/actions/runs/10268180431/job/28410799619

Passing build and push on my copy

api/Dockerfile
```diff
@@ -6,20 +6,14 @@ EXPOSE 8080
...
-ADD ./api/pyproject.toml api/pyproject.toml
-ADD bia-shared-datamodels bia-shared-datamodels
+COPY . .
 
 WORKDIR /bia-integrator/api
-RUN poetry lock
+
+# Poetry lock command likely unnecessary as we are copying the whole project
+# For local development we only need to solve the lock once
 RUN poetry install
```
liviuba, line 17: I think we also need to add poetry.lock to git if we do this. Also, even without resolving dependencies we would still pull all projects every time we build (e.g. on local). Can't we just re-add poetry.lock to git and keep the ADD shenanigans to make it go faster?

It's roughly the usecase at the bottom of this section (https://docs.docker.com/build/cache/#order-your-layers) in the docker docs

ctr26: If there's no lock file then poetry install will resolve when it installs.

We got rid of the lock file because it kept falling out of date when people make changes but didn't push lock changes.

There was a brief period of time where that was fixed with a pre-commit-hook but I don't think everyone adhered to installing that either?

ctr26: Also, git should be building with no-cache aswell to ensure it's all deterministic.

liviuba: Oh, you're right, forgot about the lockfile thing.

But can we keep the two-step add (lockfile/pyproject first, then install, then everything else on top of that)? That would make it so I don't always pull dependencies (even if they're resolved) when I rebuild on local as I'm changing code

ctr26: Ok, in my defence, I did not know you could run 

`poetry install` without providing any actual source files and that it then just magically symlinks it all later anyway when you do` poetry run `

My original answer was that you need to do ADD before poetry install so that it has your source files to install. But it doesn't, I double checked. 

Sorry for wasting time on my own baseless assumptions

ctr26:
Yeah so til you can run` poetry install ` a project with no source files, but then when you add them later poetry is happy to find them for you. This is useful for docker stuff because you can cache the lock and toml files early in the build chain.

api/Dockerfile
liviuba: This was actually a warning I was always getting but ignoring because it's a warning! Since we just need the dependencies installed and run uvicorn from the project root, I think we could do `--no-root` so `bia-integrator-api` itself gets skipped, I didn't catch that / was just thinking of it as installing dependencies, not installing the project itself

Also, should we maybe keep the comment line `#   only add poetry.lock if it exists (building on local)`? It's there because the ADD below is weird, the `poetry.lock*` is 'add if it exists', because otherwise if it's missing Docker errors

ctr26:
general info here

https://nanmu.me/en/posts/2023/quick-dockerfile-for-python-poetry-projects/

api/Dockerfile
ctr26: Add `--no-root`

ctr26:
@liviuba ready to merge

liviuba (approved):
LGTM!

PR #142, opened 2024-08-06
Add shared mock objects

liviuba:
Add shared mock objects to (hopefully) be able to reuse anywhere we need dummy data that doesn't get out of sync with the models


sherwoodf (approved):
LGTM

kbab (approved):
LGTM

PR #144, opened 2024-08-09
Fix windows path error

kbab:
Fix error due to formation of url from file paths on windows
during CI/CD using `as_posix` method of `Path`


kbab:
Fix error due to formation of url from file paths on windows
during CI/CD `as_posix` method of `Path`

sherwoodf (approved):
Lgtm - need to explore the other failing test - looks unrelated

PR #145, opened 2024-08-13
Create specimen 2

kbab:
Create bia_shared_models.Specimen and models it depends on (Biosample, SpecimenGrowthProtocol, SpecimenImagingPreparationProtocol) and update CLI to write out models that can be ingested so far.


bia-ingest-shared-models/bia_ingest_sm/conversion/specimen.py
```diff
@@ -0,0 +1,137 @@
...
+
+        model_dict = filter_model_dictionary(model_dict, bia_data_model.Specimen)
+        model_dicts.append(model_dict)
+    specimens = dicts_to_api_models(model_dicts, bia_data_model.Specimen)
+
+    if persist_artefacts and specimens:
+        persist(specimens, "specimens", submission.accno)
+
+    # ToDo: How should we deal with situation where specimens for a
+    # submission are exactly the same? E.g. see associations of S-BIAD1287
```
sherwoodf, line 116: Can you explain this a bit more? i don't follow what's wrong with this example?

sherwoodf (approved):
LGTM - left some nit-picky / suggestion comments.

bia-ingest-shared-models/bia_ingest_sm/conversion/specimen.py
kbab: For [S-BIAD1287]( https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1287 ) the associations have the following values:

```
association 1 = {"Biosample": "A549 cells", "Specimen": "A549 cells", "Image acquisition": "Label-free and light-sheet imaging"}
association 2 = {"Biosample": "Beads", "Specimen": "Beads", "Image acquisition": "Label-free and light-sheet imaging"}HEK293T cells
association 3 = {"Biosample": "HEK293T cells", "Specimen": "HEK293T cells", "Image acquisition": "Label-free and light-sheet imaging"}
association 4 = {"Biosample": "HEK293T cells", "Specimen": "HEK293T cells", "Image acquisition": "Label-free and light-sheet imaging"}
association 5 = {"Biosample": HEK293T, "Specimen": "OC43-infected HEK293T cells.", "Image acquisition": "Label-free and light-sheet imaging"}

```

`association3 == association4`. We are creating  and writing the `bia_shared_model.Specimen` twice. I am conflicted about addressing this in the code as it is an infrequent occurrence and makes logic more complex.

sherwoodf: But a) those associations are on different study components, so should be used on different datasets, and b) the biostudies specimen, biosample, and image acquisition are still unique - so assuming 1 study component has 2 associations that are equivalent we should just merge them. Do you have an example of 2 specimens with exactly the same biostudies name/id such that the association cannot be resolved?

Also, something that i now realise needs flagging - the reason Specimens were embedded nodes in Images was a) we expect everyone to skip through them to get to the actual useful information and b) technically if they're REMBI specimens (which i think we kind of assumed they should be) then they should be unique up to an Image - therefore they also need to take in an image UUID

sherwoodf: On the 2nd paragraph: i think we had an agreed upon design for the original state of the models, but then made a change to accommodate the API & didn't realise the implications - we should probably turn it into a design document to make sure we agree on that modelling decision...

kbab: A design document is a good idea. Will merge into main as is. Created clickup ticket ( https://app.clickup.com/t/8695dn7xw ) to discuss further.

sherwoodf (approved):
Approving for v1 of website - we may well need to change the logic here after the planned datamodel discussion

PR #146, opened 2024-08-13
Precommit hooks

ctr26:
- We had a go at using precommit hooks in https://github.com/BioImage-Archive/bia-agent/blob/main/.pre-commit-config.yaml and it was fine/good
- Added them here to for formatting the repo
  - N.B that you can't use black and ruff-format together as they can end up disagreeing

Also note that if you make a commit that the ruff formatter doesn't like, it'll make the changes to the file then you'll have to add the file again.

If you add a file with bad linting (unused variable) the commit won't pass and you'll have to find the issue.

Currently missing: linting and formatting the webdeveloment stuff

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,20 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v2.3.0
+    hooks:
+    -   id: check-yaml
+    -   id: end-of-file-fixer
+    -   id: trailing-whitespace
+  # - repo: https://github.com/psf/black
+  #   rev: 22.10.0
+  #   hooks:
+  #   -   id: black
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
+    rev: v0.4.8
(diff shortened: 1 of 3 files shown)


kbab (approved):
LGTM. A couple of questions though:
1. This seems to assume user has pre-commit installed globally? (I have it with pipx)
2. Will someone have to run ruff on the whole repository and make a commit?

ctr26:
1. I've left `precommit` out as a repo dependency because
a.  the hook is optional but preferred 
b.  I don't want to have to put it in each pyproject toml as a dev dep, that's hassle
c.  there's no way that I know of of getting poetry to install `precommit` and to run `precommit install` for you anyway
d. I think I can add it as a GitHub action later anyway (maybe)
3. The hook only runs per file as they're committed

kbab:
> 2. The hook only runs per file as they're committed

I asked about (2.) because I was not sure if we want a common base where all existing source is formatted by ruff - so that for new commits the only changes should be due to actual code and not difference in pre-ruff and post-ruff formatting ...

sherwoodf (approved):

PR #147, opened 2024-08-14
Ingest ImageAcquisition

kbab:
Ingest ImageAcquisition objects and write to disk. Also modify ImageAcquisition.imaging_method_name to be  a list of strings


sherwoodf (approved):
LGTM!

PR #151, opened 2024-08-19
Make directory names of ingest artefacts consistently plural

kbab:
Make all the directory names of ingested artefacts plural


sherwoodf (approved):
LGTM

PR #154, opened 2024-08-20
Api v2 pydantic v2

liviuba:
https://app.clickup.com/t/86954wmxw

Turns out Pydantic 2 was supported, it was just `generateSourceCodeOnly` in the openapi config getting in the way. Keeping the client v1 around for reference for the examples, will delete in a later PR.

Technically the only non-generated changes in this PR are the ones in openapi_config_python (toggled false to generate from scratch and then back to true to get a tidy-er client) and the Readme.md changes (openapitools/openapi-generator-cli:v7.8.0 for the generator)

Wouldn't really rely on having a "usable api client" until [this](https://app.clickup.com/t/8695873zu) is done though


sherwoodf (approved):
Exciting!

PR #155, opened 2024-08-21
Make core module use old API (archived to clients/python_v1)

kbab:
Make core module use old API (archived to clients/python_v1)

### DIFF
--- a/tools/dev/pyproject.toml
+++ b/tools/dev/pyproject.toml
@@ -15,6 +15,7 @@ typer = {extras = ["all"], version = "^0.7.0"}
 fs = "^2.4.16"
 dev_bia-integrator-core = { path = "../../core/dev", develop = true }
 bia-integrator-api = { path = "../../clients/python", develop = true }
+pydantic = "2.4.0"
 
 [build-system]
 requires = ["poetry-core"]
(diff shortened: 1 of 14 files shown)


tools/dev/pyproject.toml
```diff
@@ -15,6 +15,7 @@ typer = {extras = ["all"], version = "^0.7.0"}
 fs = "^2.4.16"
 dev_bia-integrator-core = { path = "../../core/dev", develop = true }
 bia-integrator-api = { path = "../../clients/python", develop = true }
+pydantic = "2.4.0"
```
ctr26, line 18: Is this needed if bia-integrator-api depends on pydantic already?

ctr26: My concern would be if `bia-integrator-api `updates pydantic in the future then this pinned version will clash.

ctr26:
These commits have a distinct sense of having the .pre-commit hook running. Is it working ok?

kbab:
> These commits have a distinct sense of having the .pre-commit hook running. Is it working ok?

In my opinion - yes. It also does some useful things apart from formatting, like point out unused variables.

ctr26:
Amazing! Glad it's working well for you, the linting stuff is really helpful for me too

liviuba (approved):
LGTM! Just to check, this is meant to just use the previous version of the api, not pass all tests yet, right?

PR #156, opened 2024-08-21
Manually type ambiguous links

liviuba:
For https://app.clickup.com/t/8695ft0bh

Ended up adding a `workaround_*` attribute to ObjectReference instead of adding manually (like in 1st commit) because that made referenced object validation a lot messier. It's relatively isolated / easy to change when we disambiguate the models (we can just search for the field)

Not sure if I should put the example in `test_known_bug_should_not_pass_but_does_union_reference_typed_lists_not_exclusive` down as a "known issue" or if it's worth fixing. I thing it also came up yesterday that we might have a usecase where we do actually want to mix the union types

### DIFF
--- a/api/api/models/repository.py
+++ b/api/api/models/repository.py
@@ -173,7 +173,9 @@ class Repository:
         uuid: shared_data_models.UUID,
         doc_type: Type[shared_data_models.DocumentMixin],
     ) -> Any:
-        doc = await self._get_doc_raw(uuid=uuid)
+        doc = await self._get_doc_raw(
+            uuid=uuid, model=doc_type.get_model_metadata().model_dump()
+        )
 
         if doc is None:
             raise exceptions.DocumentNotFound("Document does not exist")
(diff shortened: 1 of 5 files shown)


api/api/models/repository.py
```diff
@@ -132,13 +140,18 @@ async def assert_model_doc_dependencies_exist(
...
 
-            raise exceptions.DocumentNotFound(
-                f"Document {object_to_check.uuid} has missing dependencies: {dependencies_missing}"
-            )
+            if dependencies_missing:
+                raise exceptions.DocumentNotFound(
+                    f"Document {object_to_check.uuid} has missing dependencies: {dependencies_missing}"
+                )
+            else:
+                # all objects in the query (by uuid) were found in either one of the expected type options
```
kbab, line 153: Is the else block included for emphasis? It doesn't do anything code-wise

kbab (commented):
I looked at the use of ObjectReference in detail from the modelling part and it seems OK to me. However, I am not familiar with most of the API code and struggled to make sense of the detail. Also I could not get all the tests to run successfully on my branch.

I could not see anything wrong - happy to merge if @liviuba is happy with this level of review

api/api/models/repository.py
liviuba: yep, just emphasis to keep track of what's going on. I wanted to add it because by that point, objects can either be explicit dependencies, or the Union-workaround ones, and it was hard to keep track of what's the state when you go past the if

PR #157, opened 2024-08-21
Fix new client

liviuba:


sherwoodf (approved):

ctr26:
Does the client generation need to be part of the CI at some point in the future?

PR #158, opened 2024-08-21
Remove / fix api v1 tests failing

liviuba:
https://app.clickup.com/t/8695g4xcg

To remove the failing api tests that we have been ignoring since starting api v2 (should have probably been part of the first PR there)

So for the api now, we only run `test_minimal.py` (the tests for the new api)

### DIFF
--- a/.github/workflows/api.yaml
+++ b/.github/workflows/api.yaml
@@ -38,6 +38,7 @@ jobs:
         run: |
           docker compose up -d --build 
       - name: Run tests
+        # api moved to api/ , src left in as a reference as we transition. All tests there are failing and that directory will be deleted anyway
         run: |
           docker compose run -T bia-integrator-api poetry run pytest --ignore src
       # Always cleanup - even for cancelled jobs
--- a/.github/workflows/api.yaml
+++ b/.github/workflows/api.yaml
@@ -39,7 +39,7 @@ jobs:
           docker compose up -d --build 
       - name: Run tests
         run: |
-          docker compose run -T bia-integrator-api poetry run pytest
+          docker compose run -T bia-integrator-api poetry run pytest --ignore src
       # Always cleanup - even for cancelled jobs
       - name: Docker Compose Down
         run: |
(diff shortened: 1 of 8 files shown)


api/helm-chart/templates/deployment.yaml
```diff
@@ -25,16 +25,18 @@ spec:
...
               protocol: TCP
-          livenessProbe:
-            httpGet:
-              path: /v1/search/studies?limit=1
-              port: http
-            initialDelaySeconds: 10
-            failureThreshold: 10
-            periodSeconds: 60
+          #livenessProbe:
+          #   @TODO: Can't use a dummy uuid (404). Re-add when we have an endpoint that lists objects + pagination
```
ctr26, line 29: Thanks

.github/workflows/api.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
           docker compose up -d --build 
       - name: Run tests
         run: |
-          docker compose run -T bia-integrator-api poetry run pytest
+          docker compose run -T bia-integrator-api poetry run pytest --ignore src
```
ctr26, line 42: Would appreciate a comment on why --ignore src

ctr26:
All good!
- I can't immediately see why src is getting ignored by docker-compose
- If it's needed can it be added to .dockerignore?

liviuba:
> All good!
> 
> * I can't immediately see why src is getting ignored by docker-compose
> * If it's needed can it be added to .dockerignore?

Fixed, thanks! The only failing (api-related) test now is the one that pulls the built image. I think something changed in the meanwhile and it's not built/pushed anymore?

ctr26:
I think the ci only builds and pushes the image on changes to main? Which I think is the correct behaviour as tagged images should be immutable. Obviously this behaviour breaks the helm chart testing if it's trying to use an image that isn't built.

liviuba:
Oh ok, that makes sense, I wasn't sure what I was missing, thanks! So I think I'll un-draft this and then see if it the job still fails after merging?

ctr26:
Happy with this being merged yes

ctr26 (approved):

PR #160, opened 2024-08-21
Add hotfix. No root cause

liviuba:
Quick fix for the bug you spotted. Added https://app.clickup.com/t/8695g5yjk to check properly later

### DIFF
--- a/bia-export/bia_export/website_models.py
+++ b/bia-export/bia_export/website_models.py
@@ -1,6 +1,11 @@
 from __future__ import annotations
+
 from pydantic import Field, BaseModel
 from typing import List, Optional
+
+# Removing this breaks tests - see https://app.clickup.com/t/8695g5yjk
+from bia_shared_datamodels.bia_data_model import UUID, Annotated, ObjectReference
+
 from bia_shared_datamodels import bia_data_model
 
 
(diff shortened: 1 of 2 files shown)


bia-export/bia_export/website_models.py
```diff
@@ -1,6 +1,11 @@
 from __future__ import annotations
+
 from pydantic import Field, BaseModel
 from typing import List, Optional
+
+# Removing this breaks tests - see https://app.clickup.com/t/8695g5yjk
+from bia_shared_datamodels.bia_data_model import UUID, Annotated, ObjectReference
```
sherwoodf, line 7: So between this and the model_rebuild, i kind of lean towards the model rebuild. If we use the models elsewhere we'll have this issue, and like linkml (not that it always makes the best choices) generate models with model_rebuild() called on all classes at the end of the file - i'd kind of be more tempted to do this for all the models just in case.

liviuba: Makes sense, one thing that concerns me is that we don't know the underlying reason for the error (or at least I don't, doesn't make any sense to me). So if we add the `model_rebuild`s and make it non-obvious, it doesn't pops up periodically and won't get addressed.

Happy to delete this and just go with the model_rebuild option though, it's definitely a better long-term solution than this too if there's no better workaround, but I'd really like to understand why we get the error before we make it go away

liviuba: model_rebuild for all classes that use Annotated + a large comment seems like a better way to go. Should I change the PR do to that? (so undo the changes + add comment)

sherwoodf: I think so! Maybe leaving a comment about the errors similar to what was left for the import statement.

PR #161, opened 2024-08-22
Add auth

liviuba:
https://app.clickup.com/t/8695e0h70

Adds auth from old api (mostly moved over)


kbab (approved):
All good. @liviuba to make minor change to .env_compose and fix logs in future PR

liviuba:
> All good. @liviuba to make minor change to .env_compose and fix logs in future PR

Just realised that the readme does mention a .env, but it doesn't explicitly mention the usecase where we start the api with docker compose up and want to run the tests from the host.

In this case, what we would need to do is to `cp .env_compose .env` and change the `biaint_mongo` host in the connstring to be `localhost`. WIll add to the readme.

PR #162, opened 2024-08-22
added file counts to website models

sherwoodf:
Added file count, image count, and file-type list (using suffix, rather than biostudies format) to datasets, to display information on website.

### DIFF
--- a/bia-export/bia_export/website_conversion.py
+++ b/bia-export/bia_export/website_conversion.py
@@ -127,6 +127,22 @@ def create_experimental_imaging_datasets(
             },
         }
 
+        eid_counts_map = {}
+        file_reference_directory = root_directory.joinpath(
+            f"file_references/{accession_id}/*.json"
+        )
+        file_reference_paths = sorted(glob(str(file_reference_directory)))
+        for file_path in file_reference_paths:
+            with open(file_path, "r") as object_file:
+                object_dict = json.load(object_file)
+                file_reference = bia_data_model.FileReference(**object_dict)
+            submission_dataset = file_reference.submission_dataset_uuid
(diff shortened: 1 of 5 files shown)


bia-export/bia_export/website_conversion.py
```diff
@@ -127,6 +127,22 @@ def create_experimental_imaging_datasets(
             },
         }
 
+        eid_counts_map = {}
+        file_reference_directory = root_directory.joinpath(
+            f"file_references/{accession_id}/*.json"
```
kbab, line 132: Will this work on Windows platforms?

sherwoodf: apparently so according to the tests?

bia-export/bia_export/website_conversion.py
```diff
@@ -155,6 +175,18 @@ def process_details_section(
...
+            if eid_dict["uuid"] in eid_counts_map:
+                eid_counts_map[eid_dict["uuid"]]["file_type_aggregation"] = sorted(
+                    list(eid_counts_map[eid_dict["uuid"]]["file_type_aggregation"])
+                )
+                eid_dict = eid_dict | eid_counts_map[eid_dict["uuid"]]
+            else:
+                eid_dict = eid_dict | {
+                    "file_count": 0,
+                    "image_count": 0,
+                    "file_type_aggregation": set(),
```
kbab, line 187: does this need to be `[]`

sherwoodf:
Change the model to use a list, so it keeps order, but the code uses a set (and then converts to list) in order to get unique values.

kbab (approved):
LGTM

PR #163, opened 2024-08-23
Annotation dataset

sherwoodf:
Updated CLI to generate annotation method and image annotation datasets

Fixed file references for image annotation datasets failing to be created

Example study that now works: 

poetry run biaingest ingest S-BIAD916


kbab (commented):
All looks good and I ran on [S-BIAD1245](https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1245) (randomly chosen) and code worked.

However,  I have two comments before appending:
1. With the example I ran, some data from the Annotations section (a few from it's attributes) was not persisted
2. The tests have been extended to include the ImageAnnotationDataset and AnnotationMethod, but not the FileReferences written out for the ImageAnnotationDataset. I am happy to create a clickup ticket for this and assign it to myself so it does not stop merging this PR

kbab (commented):
The latest commits have fixed most of the missing fields - except "confidence level". Do we want to ingest this as well?

sherwoodf:
> The latest commits have fixed most of the missing fields - except "confidence level". Do we want to ingest this as well?

Looks like confidence level is not in the models, so i think it would be best to add that + ingestion in it's own pr

PR #168, opened 2024-08-27
New client examples

liviuba:
For https://app.clickup.com/t/8695873zu

Minimal examples for the only things implemented now (created/get, object reference validation, reverse links). Client/api should be usable with these.

Only `clients/python/bia_integrator_api/util.py` and `clients/python/examples/*` are not generated and the main code to review


sherwoodf (approved):
Assuming client doesn't need a deep review - approved!

clients/python/bia_integrator_api_README.md
```diff
@@ -45,15 +45,14 @@ configuration = bia_integrator_api.Configuration(
 with bia_integrator_api.ApiClient(configuration) as api_client:
     # Create an instance of the API class
     api_instance = bia_integrator_api.PrivateApi(api_client)
-    annotation_file_reference = bia_integrator_api.AnnotationFileReference() # AnnotationFileReference | 
 
     try:
-        # Create AnnotationFileReference
-        api_response = api_instance.post_annotation_file_reference(annotation_file_reference)
-        print("The response of PrivateApi->post_annotation_file_reference:\n")
+        # Example Custom Handler
```
kbab, line 50: In line 8 above should the configuration set host to localhost:8080 for this code sample to work locally?

liviuba: oh, I get the confusion but that is the generated readme. The fix for this is probably to template the whole file, which I'm not sure we want yet

The actual readme is the `Readme.md` file in that directory, will make sure to clarify the distinction in the (actual) readme

PR #169, opened 2024-08-28
refactor of study export logic 

sherwoodf:
does not change output but makes dropping in the code to handle the client easier at a later date.

General approach was to have:

create_X/ method that creates (lists of) website objects - normally independent of source (API or local) except when it would be terribly inefficient locally, so sometimes they do some pre-computing of local data that they save to the config.

retreieve_X: method that contains all the logic to return a list of API objects, either from local files (implemented) or from the API (currently not yet implemented)

Note i've made sure the tests haven't changed so that hopefully this can be reviewed smoothly.


bia-export/bia_export/website_export/studies.py
```diff
@@ -0,0 +1,39 @@
...
+from bia_export.website_export.utils import read_api_json_file
+from .website_models import Study, StudyCreationContext
+from bia_shared_datamodels import bia_data_model
+
+import logging
+
+logger = logging.getLogger("__main__." + __name__)
+
+
+def create_study(context: StudyCreationContext) -> Study:
```
liviuba, line 13: Not actually related to this line, but

Took me a while to catch on to `create_*` and `retrieve_*` being the actual important bits of this code. I think it would be more obvious / easier to spot the distinction if the code was split into `create.py` and `retrieve.py` instead of "by model", since this is (AFAICT) the main separation of concerns in this context

bia-export/bia_export/website_export/dataset_details.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def create_detail(
+    detail_object: BaseModel,
+    target_type: Type[DetailSection],
+    context: StudyCreationContext,
+):
+    detail_dict = detail_object.model_dump()
+    if detail_dict["uuid"] not in context.displayed_dataset_detail[target_type]:
+        detail_dict["default_open"] = True
```
liviuba, line 112: Is this part of the models?

liviuba (approved):
LGTM!

bia-export/bia_export/website_export/dataset_details.py
sherwoodf: Will add a description!

bia-export/bia_export/website_export/studies.py
sherwoodf: Makes sense - how about process or map for the create methods??

I'll see what the files look like split between create and retrieve. My concern would be around cyclical imports but i might be able avoid that.

PR #171, opened 2024-08-30
Create experimentally captured image

kbab:
First iteration of creating ExperimentallyCapturedImage. Two main functions

1. get_all_experimentally_captured_images - uses all FileReferences of 'easily convertible images' for all ExperimentalImagingDatasets in study
2. get_experimentally_captured_image - ExperimentalImagingDatasets and FileReference(s) have to be specified

This functionality has not been added to cli as we will probably want to refine some aspects of the ECI model and its creation process before running on all FileReferences in studies


bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
```diff
@@ -0,0 +1,186 @@
...
+            logger.warning(message)
+            continue
+
+        files_in_fl = flist_from_flist_fname(submission.accno, file_list_fname)
+        image_acquisition_title = [
+            association["image_acquisition"]
+            for association in dataset.attribute["associations"]
+        ]
+        # It would be nice to have a function like below ...
+        # image_acquisitions = get_image_acquisition_by_title(submission.accno, image_acquisition_title)
```
sherwoodf, line 72: We're currently storing these in the attribute field of the dataset - would that be easier to use?

bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
```diff
@@ -0,0 +1,186 @@
...
+def get_all_experimentally_captured_images(
+    submission: Submission, result_summary: dict, persist_artefacts=False
+) -> List[bia_data_model.ExperimentallyCapturedImage]:
+    """Return all experimentally captured images in all experimental imaging datasets"""
+
+    # TODO: Use of API will affect this! Retrieve if exists (Create otherwise?)
+    # TODO: Write function to get all objects of Type in accession_id subdir
+    image_acquisitions = None
+
+    # TODO: this is not really ingest - do we need result_summary?
```
sherwoodf, line 39: Aren't we going to e.g. send these to the API? Or do you mean because it's generating new content from what the user has submitted.

In any case i would assume that tracking the results of this would be helpful if we're doing it in bulk? But if it's a different CLI command to run on smaller sets it might not be so helpful.

bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
```diff
@@ -0,0 +1,186 @@
...
+    if not image_acquisitions:
+        # Get all image acquisitions in study
+        image_acquisitions = get_image_acquisition(
+            submission, result_summary, persist_artefacts=persist_artefacts
+        )
+
+    # Experimental Imaging Datasets in study
+    datasets = get_experimental_imaging_dataset(submission, result_summary)
```
sherwoodf, line 47: I'm guessing one of the decisions you had to make was whether to re-read/process the submission or use final study json as it appears in the API. I definitely can't tell what the pros & cons are, whatever is more practical to use is probably best.

bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
```diff
@@ -0,0 +1,186 @@
...
+    experimentally_captured_images = dicts_to_api_models(
+        model_dicts,
+        bia_data_model.ExperimentallyCapturedImage,
+        result_summary[submission.accno],
+    )
+
+    return experimentally_captured_images
+
+
+def get_experimentally_captured_image(
```
sherwoodf, line 106: I was kinda surprised this function wasn't used by the function above - am i correct in thinking it's creating "the examplar eci" for the dataset using all paths in the file list to generate the UUID? Are you not testing this behaviour because it's not actually going to be used most of the time (i'm a bit confused about it's use case)

sherwoodf (approved):
Approved, with some discussion comments.

bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
kbab: My comment above depends on how we run the ingest. If we create all ImageAcquisitions initially (as we currently do) then once the API is running we would retrieve them from the API. However, if we need to create them for any reason I agree that we should track the results.

bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
kbab: Like with the comment above I think this will be clearer once the API is in use and we settle on a workflow to use.

bia-ingest-shared-models/bia_ingest_sm/conversion/experimentally_captured_image.py
kbab: Do you mean using the ImageAcquisition title instead of its uuid? That would be easier, but from the API point of view I think the uuid makes retrieving children/dependants easier. I will mention in team meeting.

PR #174, opened 2024-09-02
Image links

sherwoodf:
Added canonical image representation* to exported image to provide information for website page

*Note that the logic to choose this image rep is pretty dodgy at the moment. It's not clear to me what the logic should be, but we should update this as start to actually process images  & therefore have a better idea of what is going on.

Added list of images to datasets to create table information for study pages

Tickets: https://app.clickup.com/t/8695jf7mt , https://app.clickup.com/t/8695hy2qk


bia-export/bia_export/website_export/website_models.py
```diff
@@ -60,24 +71,47 @@ class SpecimenImagingPreparationProtocol(
...
+    canonical_representation: bia_data_model.ImageRepresentation = Field(
+        description="""The image represenation that contains the most 'ground truth' version of the image.
+        Commonly this would be the one that captures the most data, both in terms of image metadata
+        (e.g. records of the physical dimensions of the Field of View) but also in the visual
+        fidelity of the image (e.g. has the largest field of view, or is in a data format storing the
+        greatest visual detail). This is the image represetation that should be viewable on the website."""
+    )
 
 
 class StudyCreationContext(BaseModel):
```
liviuba, line 101: The reason for it makes sense, but I'm a bit worried of this becoming an everything object, basically equivalent to globals for readability. As in, a function prototype that takes a StudyCreationContext instance would give no hints for what it does and what's the minimal set parameters it needs.

Could it be simplified sometime later when the api is up and running?

bia-export/bia_export/website_export/website_models.py
```diff
@@ -60,24 +71,47 @@ class SpecimenImagingPreparationProtocol(
...
-    acquisition_process: List[ImageAcquisition] = Field()
-    subject: Specimen = Field()
+    acquisition_process: List[ImageAcquisition] = Field(
+        description="""The processes involved in the creation of the image."""
+    )
+    subject: Specimen = Field(
+        description="""The specimen that was prepared for and captured in the field of view of the image."""
+    )
+    canonical_representation: bia_data_model.ImageRepresentation = Field(
+        description="""The image represenation that contains the most 'ground truth' version of the image.
```
liviuba, line 93: typo represenation

bia-export/bia_export/website_export/images.py
```diff
@@ -92,7 +95,13 @@ def create_image(
     }
     website_fields["subject"] = Specimen(**specimen_dict)
 
-    image_dict = api_image.model_dump() | website_fields
+    canoncical_img_rep = retrieve_canonical_representation(api_image.uuid, context)
+
+    image_dict = (
+        api_image.model_dump()
+        | website_fields
+        | {"canonical_representation": canoncical_img_rep}
```
liviuba, line 103: non-canonical spelling here

bia-export/bia_export/website_export/website_models.py
liviuba: LE: Code does look tidier with a big context object, the readability thing is just something to keep in mind. Maybe fixable with more verbose function naming?

sherwoodf: When i first wrote some of this i was essentially thinking globals but i know that they can be a pain to deal with when testing - would you prefer global variables?

It's essentially like an argparse object but with some extra stuff, so i definitely see that it could get out of hand. I would like to have a better solution, part of which might also involve changing the ingestion file structure (since this is used to store some of that information as well).

liviuba: No no, the point was exactly that it's equivalent to globals in terms of being able to tell what a function does just by its prototype, because they functions become `function_name(big_context_object)` instead of for example `function_name(study, dataset)` which would give a hint that it does something with studies and datasets.

Code looks tidy like this, but if we end up over-using it in the future we might want to factor bits out of the StudyCreateContext class

liviuba (approved):
LGTM!

PR #179, opened 2024-09-03
Add db timeout

liviuba:
https://app.clickup.com/t/8695j3phx

### DIFF
--- a/api/api/models/repository.py
+++ b/api/api/models/repository.py
@@ -19,9 +19,9 @@ from bson.binary import UuidRepresentation
 from pydantic_core import Url
 
 DB_NAME = os.environ["DB_NAME"]
+DB_TIMEOUT_MS = os.environ.get("TIMEOUT_MS", 500)
 COLLECTION_BIA_INTEGRATOR = "bia_integrator"
 COLLECTION_USERS = "users"
-COLLECTION_OME_METADATA = "ome_metadata"
 
 
 class DateCodec(TypeCodec):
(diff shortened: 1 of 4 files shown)


api/api/models/repository.py
```diff
@@ -19,9 +19,9 @@
 from pydantic_core import Url
 
 DB_NAME = os.environ["DB_NAME"]
+DB_TIMEOUT_MS = os.environ.get("TIMEOUT_MS", 500)
```
kbab, line 22: This is looking for TIMEOUT_MS, but in .env_template it says "DB_TIMEOUT_MS" - is this ok?

api/api/models/repository.py
```diff
@@ -62,25 +62,31 @@ class Repository:
...
             codec_options=CodecOptions(
                 type_registry=TypeRegistry([DateCodec(), UrlCodec()]),
                 uuid_representation=UuidRepresentation.STANDARD,
             ),
         )
-        self.users = self.db[COLLECTION_USERS]
-        self.biaint = self.db[COLLECTION_BIA_INTEGRATOR]
-        self.ome_metadata = self.db[COLLECTION_OME_METADATA]
+        self.users = self.db[collection_users]
+        self.biaint = self.db[collection_bia_integrator]
```
kbab, line 89: Will this line be added in the future? `self.ome_metadata = self.db[COLLECTION_OME_METADATA]` ? if so it may be worth just commenting it out with a 'TODO' above.

kbab (approved):
LGTM - just two comments, so approved.

api/api/models/repository.py
liviuba: not sure, depends on how we want to aggregate it, just thought I should remove it if it wasn't used, but can add it / the previous logic back if we need it

api/api/models/repository.py
liviuba: I wasn't sure if I should make it configurable in the .env or not (and cascade the change everywhere), so what this does is to make it overridable but with a default to 500ms

PR #180, opened 2024-09-03
Annotation Export

sherwoodf:
added export logic for annotation components, methods, and derived images

Also fixed if statement for images (!= should have been == to match dataset uuids)!


liviuba (approved):
LGTM!

PR #183, opened 2024-09-04
None defaults for optional annotation fields

sherwoodf:
Changed the default to None for some optional fields since we should always return 'null' in json for these and never ""


kbab (approved):
LGTM

PR #184, opened 2024-09-04
Add attribute image_representation_type to ImageRepresentation

kbab:
Add image_representation_type attribute to ImageRepresentation and update mock_objects to include a value for testing

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
@@ -31,9 +31,7 @@ class Study(ConfiguredBaseModel):
         description="""The title of a study. This will usually be displayed when search results including your data are shown."""
     )
     release_date: date = Field(description="""Date of first publication""")
-    description: str = Field(
-        description="""Brief description of the study."""
-    )
+    description: str = Field(description="""Brief description of the study.""")
     keyword: Optional[List[str]] = Field(
         default_factory=list,
         description="""Keywords or tags used to describe the subject or context of the study.""",
@@ -142,6 +140,19 @@ class LicenceType(str, Enum):
     CC_BY_SA_21_JP = "CC_BY-SA_2.1_JP"
 
 
+class ImageRepresentationType(str, Enum):
+    """Enumerate types of ImageRepresentations"""
+
+    # Original format uploaded with the study
+    UPLOADED_BY_SUBMITTER = "UPLOADED_BY_SUBMITTER"
+    # Representative image for study
+    REPRESENTATIVE = "REPRESENTATIVE"
+    # To be used as thumbnail
+    THUMBNAIL = "THUMBNAIL"
+    # In ome zarr format (usually by conversion using bioformats2raw
+    OME_ZARR = "OME_ZARR"
+
+
 #######################################################################################################
 # Subgraph 2: Contributors & their affiliations
 #######################################################################################################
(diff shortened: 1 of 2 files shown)


bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
```diff
@@ -142,6 +140,19 @@ class LicenceType(str, Enum):
     CC_BY_SA_21_JP = "CC_BY-SA_2.1_JP"
 
 
+class ImageRepresentationType(str, Enum):
+    """Enumerate types of ImageRepresentations"""
+
+    # Original format uploaded with the study
+    UPLOADED_BY_SUBMITTER = "UPLOADED_BY_SUBMITTER"
+    # Representative image for study
+    REPRESENTATIVE = "REPRESENTATIVE"
```
sherwoodf, line 149: Still not sure what representative means. If it's the representative image for the study, that sounds more like a property between the study and an image representation. 

Might 'static display' be more accurate?

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
```diff
@@ -142,6 +140,19 @@ class LicenceType(str, Enum):
...
+    # To be used as thumbnail
+    THUMBNAIL = "THUMBNAIL"
+    # In ome zarr format (usually by conversion using bioformats2raw
+    OME_ZARR = "OME_ZARR"
```
sherwoodf, line 153: Can't we figure out that something is omezarr from the format? Should this actually be something like 'interactive display' for the image pages?

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
```diff
@@ -290,10 +297,13 @@ class ImageRepresentation(ConfiguredBaseModel):
     This object was created from one or more file refences (usually one) provided by submitters to the BioImage Archive.
     """
 
-    # TODO: representation_of information is stored in representation_of_uuid defined in bia_data_model. 
+    # TODO: representation_of information is stored in representation_of_uuid defined in bia_data_model.
     # The field here will eventually be used to generate endpoints.
     # representation_of: AbstractImageMixin = Field( description="The abstraction of this image represtation.")
     image_format: str = Field(description="""Image format of the combined files.""")
+    image_representation_type: ImageRepresentationType = Field(
```
sherwoodf, line 304: I suspect this should be 'use_type' because really it's hints as to what we expect the image rep to be used for.

Will you be updating tests in a subsequent PR? If you want we can split up fixing the different repositories if that would make it easier.

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
kbab: When we display a study, the 'representative image' is used on the study page. I will change it to 'STATIC_DISPLAY', but we can modify it in a later iteration if necessary

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
kbab: Yes - the suffix should be .ome.zarr. Will change to INTERACTIVE_DISPLAY

bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
kbab: Thanks - will change to 'use_type'

PR #186, opened 2024-09-05
Export img dataset

sherwoodf:
Updates that:

- Fix test from model updates.
- Generates metadata for datasets.
- Split up some of the CLI options to keep the fields relevant to the particular command being run.


bia-export/bia_export/cli.py
```diff
@@ -4,8 +4,11 @@
 from rich.logging import RichHandler
 from typing_extensions import Annotated
 from pathlib import Path
-from .website_export.study_pages.transform import transform_study
-from .website_export.image_pages.transform import transform_ec_images
+from .website_export.studies.transform import transform_study
+from .website_export.studies.models import CLIContext as StudyCLIContext
+from .website_export.images.transform import transform_ec_images
+from .website_export.images.models import CLIContext as ImageCLIContext
```
liviuba, line 10: I would avoid naming things the same if they're used a lot to avoid confusion. Ex there is some CLIContext below (for retrieve_datasets) which I can't easily tell if it's a study or image one, and it wouldn't be unreasonable to be either one

liviuba (changes requested):
Minor naming change to make code less confusing. It might be better if you add the client sooner rather than later, to avoid a big refactor for Context?

bia-export/bia_export/cli.py
sherwoodf: Yeah i wasn't a fan but was rushing and this was easier. Fixed.

liviuba (approved):
LGTM! Looking forward to seeing the api client in use!

PR #187, opened 2024-09-05
Fix db params exposed through api

liviuba:
https://app.clickup.com/t/8695mu9k9

Looks like get_db defined previously was never actually used, and fastapi was DI-ing a new instance of Repository (-> the exposed params). Changed how we request db instances in handlers + small change&comment to Repository to make this harder to misconfigure in the future


AybukeKY (approved):
LGTM

PR #190, opened 2024-09-06
added api client version of the exporter

sherwoodf:
https://app.clickup.com/t/86956nun1

Adds client integration for the exporter

Minor update to cli method, such as type hints

Only export of study was tested - waiting on regenerated client with some test data to manually test.


liviuba (approved):
LGTM! Do you still need the test data in the api?

PR #193, opened 2024-09-06
Export to use Client Models

sherwoodf:
Pydantic was compaining when trying to create objects due to a model mis-match, despite them functionally being the same e.g.
```
class Specimen (BaseModel)
   sample_of: List[bia_shared_models.BioSample] = Field(
        description="""The biological matter that sampled to create the specimen."""
    )
```
will fail validation for a dictionary of the form:
```
{
  sample_of: [
     (api_client.BioSample( ...))
  ]
}
```

I switched to using client models as much as possible (except for a few type hits where i deal with higher level classes that aren't in the clients).

I could have instead called .model_dump() in a bunch of places, but i'm not sure that's that helpful, and i think it's better that the file version of the export code uses the same models as the api version.

Changes ended up added the default_open to the image models, which is fine - it's not strictly needed - i would rather find a longer term solution in Astro for working out which details sections have been displayed & get rid of this extra field everywhere. 

Also fixed copy-paste errors that came up with Annotated vs Experimental related differences.

### DIFF
--- a/bia-export/bia_export/website_export/datasets_for_images/models.py
+++ b/bia-export/bia_export/website_export/datasets_for_images/models.py
@@ -1,12 +1,8 @@
 from __future__ import annotations
 
-from bia_shared_datamodels import bia_data_model
+from bia_integrator_api import models
 from pydantic import Field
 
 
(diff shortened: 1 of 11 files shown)


liviuba (approved):
LGTM!

PR #194, opened 2024-09-06
Image conversion

kbab:
First steps towards image conversion. Create command line tool to accept accession ID and uuid of file reference and convert to zarr and update image representations then save to disk

Most of the functions in the image_utils modules are needed temporarily whilst bia_converter is being developed, so some of them do not have type hints, and due to time pressures and the fact that they are temporary, these were not corrected.

### DIFF
--- a/bia-ingest-shared-models/bia_ingest_sm/config.py
+++ b/bia-ingest-shared-models/bia_ingest_sm/config.py
@@ -18,6 +18,11 @@ class Settings(BaseSettings):
     )
 
     bia_data_dir: str = Field(default_output_base)
+    endpoint_url: str = Field("https://uk1s3.embassy.ebi.ac.uk")
+    bucket_name: str = Field("bia-integrator-data")
+    cache_root_dirpath: Path = Field(Path.home() / ".cache" / "bia-converter")
+    bioformats2raw_java_home: str = Field("")
+    bioformats2raw_bin: str = Field("")
 
 
 # class Settings:
--- a/bia-ingest-shared-models/bia_ingest_sm/conversion/image_representation.py
+++ b/bia-ingest-shared-models/bia_ingest_sm/conversion/image_representation.py
@@ -46,6 +46,7 @@ def create_image_representation(
 
     pixel_metadata = {}
     total_size_in_bytes = 0
+    file_uri = []
     if image_format == ".ome.zarr":
         pixel_metadata = image_utils.get_ome_zarr_pixel_metadata(
             representation_location
@@ -58,11 +59,12 @@ def create_image_representation(
         # TODO: Discuss what to do if file reference is list > 1 with different paths e.g. .hdr and .img for analyze. Currently just using first file reference
         if image_format == "":
             image_format = image_utils.get_image_extension(file_references[0].file_path)
+        file_uri = [fr.uri for fr in file_references]
 
     model_dict = {
         "image_format": image_format,
         "use_type": representation_use_type,
-        "file_uri": [],
+        "file_uri": file_uri,
         "original_file_reference_uuid": file_reference_uuids,
         "representation_of_uuid": experimentally_captured_image.uuid,
         "total_size_in_bytes": total_size_in_bytes,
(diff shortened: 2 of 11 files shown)


bia-ingest-shared-models/bia_ingest_sm/config.py
```diff
@@ -18,6 +18,11 @@ class Settings(BaseSettings):
     )
 
     bia_data_dir: str = Field(default_output_base)
+    endpoint_url: str = Field("https://uk1s3.embassy.ebi.ac.uk")
```
kbab, line 21: Should we adopt the same convention as @liviuba uses for the environment variables for the API?

bia-ingest-shared-models/bia_ingest_sm/conversion/image_representation.py
```diff
@@ -93,6 +99,142 @@ def create_image_representation(
     # return dicts_to_api_models([model_dict,], bia_data_model.ImageRepresentation )[0]
 
 
+def create_images_and_image_representations(
```
kbab, line 102: Was not really sure where to place this as it bridges between creating models for API and converting/creating actual images. When we start using API this will probably be in bia_converter package.

bia-ingest-shared-models/bia_ingest_sm/image_utils/conversion.py
```diff
@@ -0,0 +1,46 @@
+# All code in this module originate from bia-converter/bia_converter/io.py
```
kbab, line 1: This was mostly copy and paste from bia_converter. It is temporary until bia_converter is fixed for API

bia-ingest-shared-models/bia_ingest_sm/image_utils/io.py
```diff
@@ -0,0 +1,116 @@
+# All code in this module originate from bia-converter/bia_converter/io.py
```
kbab, line 1: This was mostly copy and paste from bia_converter. It is temporary until bia_converter is fixed for API

bia-ingest-shared-models/bia_ingest_sm/image_utils/omezarrmeta.py
```diff
@@ -0,0 +1,67 @@
+from typing import List, Optional
```
kbab, line 1: This was mostly copy and paste from bia_converter. It is temporary until bia_converter is fixed for API

bia-ingest-shared-models/bia_ingest_sm/image_utils/rendering.py
```diff
@@ -0,0 +1,379 @@
+from typing import Dict, List, Optional
```
kbab, line 1: This was mostly copy and paste from bia_converter. It is temporary until bia_converter is fixed for API

bia-ingest-shared-models/pyproject.toml
```diff
@@ -15,6 +15,12 @@ bia-shared-datamodels = { path = "../bia-shared-datamodels", develop = true }
 typer = "^0.12.3"
 typing-extensions = "^4.12.2"
 pydantic-settings = "^2.3.4"
+zarr = "^2.18.3"
```
kbab, line 18: These are in bia_converter, and so will be removed once bia_converter starts being used and is a requirement of this package

bia-ingest-shared-models/bia_ingest_sm/config.py
liviuba: not sure I get the question, what convention?

bia-ingest-shared-models/pyproject.toml
liviuba: A comment like `# TEMPORARY DEPENDENCIES FROM bia_converter` with the temporary dependencies below that would IMO it easier to undo the workarounds. Even after the code would get removed, it's usually hard to tell if a dependency is/isn't needed anymore, so dependency lists tend to only grow

Oh or if there aren't any conflicts, you can just add bia_converter itself as a dependency

kbab: Thanks - inserted comment as suggested

PR #195, opened 2024-09-09
Clarify docs

liviuba:
Minor changes to client docs, `get_studies` was missing

Removed placeholder example custom endpoint (because we have `get_studies` now)


AybukeKY (approved):
LGTM

PR #198, opened 2024-09-11
disk to api script speedups and cleanup

liviuba:
https://app.clickup.com/t/8695rqgx9


clients/python/example/disk_to_api_push.py
```diff
@@ -0,0 +1,160 @@
...
+    return data_parsed
+
+
+def api_push_json(delete_after_push: bool, json_path: str):
+    report = {"already_exists": [], "not_found": []}
+
+    with open(json_path) as f:
+        data_parsed = dict_parse_unknown_model(json.load(f), file_name=json_path)
+
+        try:
```
kbab, line 57: I was wondering if you are always going to stick to the convention of making the creation function post_(snake case version of model.type_name) i.e. ThisModel -> post_this_model.

If so a possible alternative to the if statements below is:
```
api_creation_method = re.split(r"(?=[A-Z])", data_parsed.model.type_name)
getattr(api_client, api_creation_method)(data_parsed)
```

kbab (approved):
LGTM - made one comment but approved

clients/python/example/disk_to_api_push.py
liviuba: the naming convention for create methods will probably stay, but the big if/else chain seemed simpler than coming up with/testing something that converts to snake_case (ex that still needs a tolower/.join('_') and would fail in a confusing way if type_name has a typo) + this looked like a quick to read reference for creation methods

but I can use to_snake from pydantic since we depend on it anyway. I'd rather add it later as a generic create function on the client though

PR #199, opened 2024-09-12
Update process to generate ExperimentallyCapturedImage

kbab:
* Use FileReference.file_path in creation of UUID for ECI to prevent errors when running tests on windows platform
* Add attributes from FileReference to ECI attributes

### DIFF
--- a/bia-ingest/bia_ingest/conversion/utils.py
+++ b/bia-ingest/bia_ingest/conversion/utils.py
@@ -269,3 +269,30 @@ def get_bia_data_model_by_uuid(
         Path(settings.bia_data_dir) / model_subdir / accession_id / f"{uuid}.json"
     )
     return model_class.model_validate_json(input_path.read_text())
+
+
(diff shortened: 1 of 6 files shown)


bia-ingest/bia_ingest/conversion/utils.py
```diff
@@ -269,3 +269,30 @@ def get_bia_data_model_by_uuid(
...
+    """
+
+    if not dict_list:
+        return {}
+
+    merged_dict = dict_list[0]
+
+    for dictionary in dict_list[1:]:
+        for key, value in dictionary.items():
+            # If the key already exists in the merged dictionary
```
liviuba, line 287: +1 for the comments, makes it much easier to track

liviuba, line 287: +1 for the comments, makes it much easier to track

liviuba (approved):
LGTM!

PR #200, opened 2024-09-13
Create one bia_model_data.Specimen object per dataset

kbab:
Following discussion with ﻿@Francois we agreed to change the format of associating SpecimenImagingPreparationProtocol, SpecimenGrowthProtocol and Biosample with a BIA Specimen object. 

See [https://app.clickup.com/t/8695fqxpy](https://app.clickup.com/t/8695fqxpy)

The current method creates one Specimen object per unique association in each study component (i.e. 3 unique associations ➝. 3 Specimens)

This PR modified it to create only one Specimen per dataset (study component) irrespective of the number of unique associations in the dataset. The fields of the Specimen object will contain lists of uuids to the various sub-objects.

### DIFF
--- a/bia-ingest/bia_ingest/conversion/specimen.py
+++ b/bia-ingest/bia_ingest/conversion/specimen.py
@@ -24,50 +24,45 @@ from . import (
 logger = logging.getLogger("__main__." + __name__)
 
 
-# TODO: Should we make this per image?
-# TODO: Need to change according to https://app.clickup.com/t/8695fqxpy
-def get_specimen_for_association(
-    submission: Submission, association: Dict[str, str], result_summary: dict
+def get_specimen_for_dataset(
+    submission: Submission,
+    dataset: bia_data_model.ExperimentalImagingDataset,
+    result_summary: dict,
 ) -> bia_data_model.Specimen:
     """Return bia_data_model.Specimen for a particular dataset"""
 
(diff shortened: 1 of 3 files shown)


bia-ingest/bia_ingest/conversion/specimen.py
```diff
@@ -24,50 +24,45 @@
...
+    dataset: bia_data_model.ExperimentalImagingDataset,
+    result_summary: dict,
 ) -> bia_data_model.Specimen:
     """Return bia_data_model.Specimen for a particular dataset"""
 
-    specimen_title = association["specimen"]
+    # According to https://app.clickup.com/t/8695fqxpy we want one specimen
+    # per dataset, so if more than one association we are concatenation
+    # the required information from each.
+    associations = [association for association in dataset.attribute["associations"]]
```
liviuba, line 37: just to check if I get understand the process, was this (associations as a dataset attribute) what we needed to create ExperimentallyCapturedImages to connect ExperimentalImagingDatasets and Specimens?

liviuba (approved):
LGTM!

bia-ingest/bia_ingest/conversion/specimen.py
kbab: No - these are the associations we capture from the Submission (associations section of study component).

PR #204, opened 2024-09-18
Image conversion using api

kbab:
Convert images to zarr format and create STATIC AND INTERACTIVE_DISPLAY images
Upload to S3
Update image_representation models in API (or on disk)
Update README and .env_template

Note that ingest does not currently write into API - only experimentally captured images and image representations. Making ingest write to API will be done in a separate PR


bia-ingest/bia_ingest/serialiser.py
```diff
@@ -0,0 +1,135 @@
+from uuid import UUID
```
liviuba, line 1: I would really not call this (any any related things) a serialiser

bia-ingest/bia_ingest/serialiser.py
```diff
@@ -0,0 +1,135 @@
...
+        for uuid in uuids:
+            input_path = (
+                self.output_dir_base / model_subdir / self.accession_id / f"{uuid}.json"
+            )
+            object_list.append(model_class.model_validate_json(input_path.read_text()))
+        return object_list
+
+
+# Serialise to the MongoDB using API
+class MongodbSerialiser(Serialiser):
```
liviuba, line 73: Is mongodb the thing this directly interacts with?

liviuba (changes requested):
Main change request is around some misleading naming, not the approach.

A minor thing to consider but happy to merge without is that generic/untyped gets like in the Serialiser* classes make IDE attribute completion support impossible and cancel the benefit of the api client explicitly being typed. Might slow down future work by having to constantly refer to the model docs / memorise them

bia-ingest/bia_ingest/serialiser.py
kbab: Happy to change the name - what do you suggest?

kbab:
> A minor thing to consider but happy to merge without is that generic/untyped gets like in the Serialiser* classes make IDE attribute completion support impossible and cancel the benefit of the api client explicitly being typed. Might slow down future work by having to constantly refer to the model docs / memorise them

Is there a way to work around this? The class was created in an attempt to allow using either the API or writing to disk . If we are really sure we do not want to have writing to disk functionality, I can use the changes you implemented in the `persist` function instead

bia-ingest/bia_ingest/serialiser.py
liviuba: I'm not sure, what makes this a serialiser as opposed to something else?

kbab: I used the name 'serialiser' because it takes an object and stores it as a json document either on disk or in the database (I assume this is serialisation?) and retrieves from json doc into object (deserialisation?) ...

liviuba:
> > A minor thing to consider but happy to merge without is that generic/untyped gets like in the Serialiser* classes make IDE attribute completion support impossible and cancel the benefit of the api client explicitly being typed. Might slow down future work by having to constantly refer to the model docs / memorise them
> 
> Is there a way to work around this? The class was created in an attempt to allow using either the API or writing to disk . If we are really sure we do not want to have writing to disk functionality, I can use the changes you implemented in the `persist` function instead

I don't think the changes I made added / were intended to add typing

bia-ingest/bia_ingest/serialiser.py
liviuba: This seems to be doing more than serialisation, and it's not serialisation format that it abstracts

ex https://en.wikipedia.org/wiki/Serialization https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats

kbab: Just looked up definition of serialisation. Is your point that it is the process of conversion to the form that can be stored as opposed to implementing the actual storage/retrieval?

kbab: > This seems to be doing more than serialisation, and it's not serialisation format that it abstracts
> 
> ex https://en.wikipedia.org/wiki/Serialization https://en.wikipedia.org/wiki/Comparison_of_data-serialization_formats

Yes it is!

kbab: How about `PersistenceStrategy` -> `DiskPersistence` and `MongodbPersistence`?

liviuba: Seems closer aligned with what it does to me. Is Mongo a direct dependency of that part of the code?

kbab: No, I guess `ApiPersistence` is more appropriate?

liviuba: yes IMO

https://en.wikipedia.org/wiki/Law_of_Demeter

sherwoodf (approved):

PR #206, opened 2024-09-24
Fix ingest errors

sherwoodf:
Pydantic requires a default value to be set for a field in order for it to actually be optional.

When BioStudies returns file lists they can return file objects with an attribute that has a name but no value key if the column is empty.

Updated the tests to have an example of this, and updated bia_ingest_cli to use a fixture so that pydantic doesn't error when loading the tests, only a run time.

According to the results spreadsheet, this might allow us to ingest another ~40-90 studies.


kbab (approved):
LGTM

PR #207, opened 2024-09-24
fix tests in github run

sherwoodf:
> tests were not failing locally, but were failing in the github actions run. Trying to change the logic to to see if the field order is different in some objects.

The issue was that i'd not run poetry update locally, and the client gets imported as an external dependency (and had been updated)

Text expectations were updated separately (oops) I still think this change will help make the tests less brittle.


kbab (approved):
LGTM

PR #208, opened 2024-09-24
Decouple creation of image reps from image conversion

kbab:
Separate creation of image representations from conversion/creation of actual images for the image representations. A new subpackage of bia-integrator (bia-integrator-light) has been created for this. The name reflects its intended use as a lightweight temporary version of bia-converter.

* Move contents of image_utils from bia-ingest/bia_ingest into bia-converter-light/bia_converter_light
* Create cli.py in bia-converter-light/bia_converter_light
* Move scripts for helping with selecting file references to convert to bia-converter-light/scripts


bia-converter-light/bia_converter_light/cli.py
```diff
@@ -0,0 +1,192 @@
...
+
+logger = logging.getLogger()
+
+
+@app.command(help="Convert image for given image representation UUID")
+def convert_image(
+    image_representation_uuid: Annotated[str, typer.Argument()],
+    verbose: Annotated[bool, typer.Option("-v")] = False,
+) -> None:
+    """Create image for supplied image rep, upload to s3 and return uri
```
sherwoodf, line 30: Also updates the image representation in the API

bia-ingest/bia_ingest/cli.py
```diff
@@ -104,6 +103,13 @@ def create(
     persistence_mode: Annotated[
         PersistenceMode, typer.Option(case_sensitive=False)
     ] = PersistenceMode.api,
+    reps_to_create: Annotated[
+        List[ImageRepresentationUseType], typer.Option(case_sensitive=False)
+    ] = [
+        ImageRepresentationUseType.UPLOADED_BY_SUBMITTER,
```
sherwoodf, line 109: Would we worth discussing after SAB feedback - but i wonder whether we might want to create all the UPLOADED_BY_SUBMITTER image reps for all the images (e.g. to logically group together the files that were uploaded), and only convert a smaller number to ZARR.

bia-ingest/bia_ingest/conversion/utils.py
```diff
@@ -307,3 +307,32 @@ def merge_dicts(dict_list: List[Dict[str, str]]) -> Dict:
                 merged_dict[key] = value
 
     return merged_dict
+
+
+# Copied over from bia-converter/bia_converter/utils.py to prevent
```
sherwoodf, line 312: Would it be easier to import to/from the bia-converter-light?

I know we want some separation of concerns between the two processes - but i would assume that that's more about the how we run it than requiring complete separation of code base. Do the lists here and in bia-converter-light need to be the same for things to run without issues? If so i'd recommend importing one from the other.

sherwoodf (changes requested):
Minor changes requested, mostly to text for other users.

I'm also a bit confused about the naming/context of this package: the 'light' part - why is this more lightweight or temporary compared to the other package?

kbab:
> Minor changes requested, mostly to text for other users.
> 
> I'm also a bit confused about the naming/context of this package: the 'light' part - why is this more lightweight or temporary compared to the other package?

This package only deals with the case of conversions of one file reference per image representation. An image representation can have more than one file reference (we will implement this is Sprint 5 starting 26/09). However, bia-converter-light will only convert the first file reference in the list, whereas bia-converter will be able to do more complex conversion including creation of multichannel images and multi-slice images from multiple file references. We can call it bia-converter, but I wanted the name to emphasise its limited capabilities.

bia-converter-light/pyproject.toml
```diff
@@ -0,0 +1,39 @@
...
+requests = "^2.31.0"
+pytest = "^7.0"
+bia-shared-datamodels = { path = "../bia-shared-datamodels", develop = true }
+bia-integrator-api = { path = "../clients/python", develop = true }
+bia-ingest = { path = "../bia-ingest", develop = true }
+typer = "^0.12.3"
+typing-extensions = "^4.12.2"
+pydantic-settings = "^2.3.4"
+
+# TEMPORARY DEPENDENCIES FROM bia_converter
```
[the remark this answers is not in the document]
kbab: I have modified the comment. It was left over when I copied the bia-ingest pyproject.toml.

bia-ingest/bia_ingest/cli.py
kbab: +1 for a discussion. I think MH and AY also suggested something similar.

bia-ingest/bia_ingest/conversion/utils.py
```diff
@@ -307,3 +307,32 @@ def merge_dicts(dict_list: List[Dict[str, str]]) -> Dict:
...
+        ".ome.tiff": ".ome.tiff",
+        ".ome.tif": ".ome.tiff",
+        ".tar.gz": ".tar.gz",
+    }
+
+    for special_ext, mapped_value in special_cases.items():
+        if file_path.lower().endswith(special_ext):
+            return mapped_value
+
+    ext_map = {
```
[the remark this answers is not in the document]
kbab: It could be incorporated into the previous map, but it serves a different purpose. I have put in comments and changed name of the first map to make logic clearer. I am also happy to combine the two maps if necessary.

bia-ingest/bia_ingest/conversion/utils.py
kbab: Changed so that bia-converter-light imports the function from bia-ingest. So bia-ingest does not depend on bia-converter-light.

bia-converter-light/README.md
```diff
@@ -0,0 +1,19 @@
...
+$ poetry run bia-converter-light convert-image da612702-e612-4891-b440-816d0a2b15be
+```
+This assumes that the UUID supplied (`da612702-e612-4891-b440-816d0a2b15be`) is that of an image representation in the API
+
+The API URI default is 'localhost:8080'. This can be changed by using an
+environment variable `BIA_API_BASEPATH` or setting the value in a `.env` file in the `bia-converter-light` directory.
+
+The base directory defaults to `~/.cache/bia-integrator-data-sm`. This can be changed by setting `bia_data_dir` environment variable, or creating a .env file in this folder setting this value.
+
+## convert-images dependencies
```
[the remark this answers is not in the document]
kbab: Happy to chat. A further goal is to containerise the commands so user can simply run a container without worrying about installation details. Can we deal with this separate from this PR?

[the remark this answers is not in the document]
sherwoodf: Yep, definitely, this can be separate.

sherwoodf (approved):

PR #209, opened 2024-09-30
Add yaml containing file reference uuids of images converted for SAB

kbab:
Add a yaml file containing file reference uuids and other details for image representations converted for the SAB


kbab:
Does the location of the yaml seem sensible? If not can a more appropriate location be suggested?

sherwoodf (approved):
LGTM!

PR #211, opened 2024-10-01
Print summary of models created/errors when creating image representations

kbab:
This PR attempts to print a summary of models created and/or errors encountered when creating image representations. However, the output generated is not strictly correct - as only one Experimentally Captured Image (ECI) is created for all representations of the same file reference uuid(s). However, the current code logs creation of an ECI for each representation - thus overstating the number of ECIs created.

### DIFF
--- a/bia-ingest/bia_ingest/cli_logging.py
+++ b/bia-ingest/bia_ingest/cli_logging.py
@@ -48,6 +48,8 @@ class IngestionResult(BaseModel):
     Organisation_ValidationErrorCount: int = Field(default=0)
     ExperimentallyCapturedImage_CreationCount: int = Field(default=0)
     ExperimentallyCapturedImage_ValidationErrorCount: int = Field(default=0)
+    ImageRepresentation_CreationCount: int = Field(default=0)
+    ImageRepresentation_ValidationErrorCount: int = Field(default=0)
 
 
 def tabulate_errors(dict_of_results: dict[str, IngestionResult]) -> Table:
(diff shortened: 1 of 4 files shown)


bia-ingest/bia_ingest/cli_logging.py
```diff
@@ -60,21 +62,24 @@ def tabulate_errors(dict_of_results: dict[str, IngestionResult]) -> Table:
                 error_message += f"{field}: {value}; "
 
         if (
-            result.ExperimentalImagingDataset_CreationCount == 0 
-            & result.ImageAnnotationDataset_CreationCount == 0
+            result.ExperimentalImagingDataset_CreationCount
```
sherwoodf, line 65: This has been fixed in a different PR already and will cause a merge conflict

bia-ingest/bia_ingest/conversion/experimentally_captured_image.py
```diff
@@ -52,16 +50,21 @@ def get_experimentally_captured_image(
         attribute=attributes,
     )
 
-    experimentally_captured_image = (
-        bia_data_model.ExperimentallyCapturedImage.model_validate(model_dict)
+    experimentally_captured_image = dicts_to_api_models(
```
sherwoodf, line 53: We have try/except blocks for creating objects with model_validate elsewhere (e.g. https://github.com/BioImage-Archive/bia-integrator/blob/main/bia-ingest/bia_ingest/conversion/experimental_imaging_dataset.py#L125). I think would read more clearly than calling the dicts method followed by an assert that it's length 1 - the current assert error wouldn't be caught and so i would expect the process to just stop?

sherwoodf (changes requested):
Two small comments on these changes, but otherwise would be okay with approving this, despite the issue with the miscounted ECIs....

I have more fundamental issues with how this part of the cli is set up (which is think the miscounting is a symptom of), but i think these should be addressed in a separate ticket to refactor the code. Namely:

1. The cli command seems to be set up differently to the ingest command, as if it will be about creating various different objects, but then the logic is only (currently) very specific to image representations - and i don't see us needing it for other things.
2. Despite being all about creating representations it also is the only way to create the abstract images, which isn't obvious from the cli code.

bia-ingest/bia_ingest/cli_logging.py
sherwoodf: It was merged last wednesday.

sherwoodf: Nvm, it's just a formatting change, not a code change.

bia-ingest/bia_ingest/conversion/experimentally_captured_image.py
kbab: Thanks - I have changed the code to use these

kbab:
> Two small comments on these changes, but otherwise would be okay with approving this, despite the issue with the miscounted ECIs....
> 
> I have more fundamental issues with how this part of the cli is set up (which is think the miscounting is a symptom of), but i think these should be addressed in a separate ticket to refactor the code. Namely:
> 
> 1. The cli command seems to be set up differently to the ingest command, as if it will be about creating various different objects, but then the logic is only (currently) very specific to image representations - and i don't see us needing it for other things.
> 2. Despite being all about creating representations it also is the only way to create the abstract images, which isn't obvious from the cli code.

Happy to discuss and create new clickup ticket to address these.

PR #212, opened 2024-10-01
Fix - to _ for the db that already existed

liviuba:
https://app.clickup.com/t/86960wc13

Existing api was deployed with different scripts with DB_NAME=bia_integrator (note -). Helm chart defaulted to bia-integrator; changed to the one with _


liviuba:
@ctr26 Fixed the failing api test, I think this is ready for review (only fixes the api pod itself, still looking into the service problem)

api/.env_compose
```diff
@@ -3,4 +3,5 @@ MONGO_INITDB_ROOT_USERNAME=root
 MONGO_INITDB_ROOT_PASSWORD=example
 DB_NAME=bia_integrator
 JWT_SECRET_KEY="0123456789=="
-USER_CREATE_SECRET_TOKEN="0123456789=="
\ No newline at end of file
+USER_CREATE_SECRET_TOKEN="0123456789=="
+DB_TIMEOUT_MS=5000
```
ctr26, line 7: good

api/Readme.md
```diff
@@ -99,9 +99,9 @@ To rebuild the api image and push it to the docker container registry.
 
 # building the api image
 docker login ghcr.io
-docker build -t bioimage-archive/integrator-api:0.1 .
-docker image tag bioimage-archive/integrator-api:0.1 ghcr.io/bioimage-archive/integrator-api:0.1
-docker push ghcr.io/bioimage-archive/integrator-api:0.1
+docker build -f api/Dockerfile -t bioimage-archive/integrator-api:0.2 .
```
ctr26, line 102: probably not quite right if someone runs this command from the api directory rather than the root_dir right?

Maybe easier to replace it with the docker-compose build commands?

PR #213, opened 2024-10-02
Add reverse link indexes

liviuba:
For https://app.clickup.com/t/8695e0ecp with a max query time config variable in an older ticket should make sure we either get fast indexed responses or are aware we have an unindexed query (by the request failing)

Had to add another env var, should probably prioritise https://app.clickup.com/t/8695q6qjd soon

To test:
* Set `DB_INDEX_REFRESH=True` in `.env_compose`
* Start the api `docker compose up -d --build`
* Check `Indexes` for the bia_integrator collection, there should be lots of indexes named similar to `imagerepresentation_original_file_reference_uuid`  (one per `*_uuid` link, like the "reverse links" in the api)


api/.env_compose
```diff
@@ -4,4 +4,5 @@ MONGO_INITDB_ROOT_PASSWORD=example
 DB_NAME=bia_integrator
 JWT_SECRET_KEY="0123456789=="
 USER_CREATE_SECRET_TOKEN="0123456789=="
-DB_TIMEOUT_MS=5000
\ No newline at end of file
+DB_INDEX_REFRESH=False
```
kbab, line 7: This is set to `False`, but user is told to change it to `True`. This leaves git repo in a 'modified' state and user has to either commit or stash if changing branches. Is there a reason why it is set to `False` here?

api/.env_template
```diff
@@ -7,4 +7,5 @@ DB_TIMEOUT_MS=500
 # run: 'openssl rand -hex 32' to generate a secret key for testing
 JWT_SECRET_KEY="long_key"
 # token needs to end in ==
-USER_CREATE_SECRET_TOKEN="some_long_secret_token"
\ No newline at end of file
+USER_CREATE_SECRET_TOKEN="some_long_secret_token"
+DB_INDEX_REFRESH=False
```
kbab, line 11: With reference to my comment about .env_compose - it makes sense if the value for DB_INDEX_REFRESH here overrides that in .env_compose. In any case I think it would be helpful if this info about DB_INDEX_REFRESH is in the Readme.md file.

api/api/app.py
```diff
@@ -24,12 +22,17 @@
 
 
 @app.on_event("startup")
-def on_start():
+async def on_start():
+    if os.getenv("DB_INDEX_REFRESH") == "True":
+        # just so "False" / other values don't accidentally trigger indexing
+        log_info("App updating indexes")
+        await repository_create(init=True)
```
kbab, line 29: Is this why setting `DB_INDEX_REFRESH=True` erases old data?

liviuba: No, nothing in this PR should erase or affect old data, not even indexes (intentionally in case we have manually added ones)

How could old data be erased here?

kbab (commented):
Works on my local version. I had an error when trying to run the api container before the DB_TIMEOUT_MS was added to .env_compose.

Happy to approve, but I think the following should be considered:
1. Updating the Readme
2. In pyproject.toml `readme = "README.md"`, but on disk the file is named `Readme.md` . It is worth making both the same.
3. Default value of DB_INDEX_REFRESH in .env_compose

api/.env_template
liviuba: Agree that DB_INDEX_REFRESH should be documented, but I'd rather add docs to .env_template instead of Readme because it's closer to where it would be changed, adding it now

api/api/app.py
kbab: Not sure. I asked because I do not know what `repository_create(init=True)` does, especially the `init=True` part.

liviuba: Oh, old naming from when I wasn't sure what it should do but knew I wanted different code to run when initialising. Renaming now, wasn't even reading the name anymore so didn't spot that it's inconsistent

liviuba:
Added https://app.clickup.com/t/86962ccu2 for the error (DB_TIMEOUT_MS of 500ms hit if DB_INDEX_REFRESH=True for a local db with data already in it) + generally better handling of index changes

1. Added docs for DB_INDEX_REFRESH in `.env_template`
2. Changed pyproject.toml
3. I think actual change is in [this](https://github.com/BioImage-Archive/bia-integrator/pull/213#discussion_r1786486243) thread, renamed `init`

kbab (approved):
LGTM

PR #214, opened 2024-10-02
Make ingest use persistence strategy (API/disk)

kbab:
Address clickup ticket https://app.clickup.com/t/8695e2y7p . The ingest command can now either persist using the API or read/write artefacts to disk. The readme has been updated.


sherwoodf:
Should the default be on disk? I'm a bit worried about the default making changing to the API...

sherwoodf:
Minor comment, but i think it would be better if the 'written' log lines were debug level - these can always be viewed with a -v, and would reduce the quantity of output when dealing with long file lists.
image

bia-converter-light/README.md
```diff
@@ -1,24 +1,45 @@
...
+* AWS_ACCESS_KEY_ID *and* AWS_SECRET_ACCESS_KEY
+<br>OR
+* AWS_SHARED_CREDENTIALS_FILE with optional AWS_PROFILE and/or AWS_CONFIG_FILE
 ## Usage
-Once you've installed the project using poetry, assuming you are in this directory:
+Once the project is installed and the environment configured, assuming this is the working directory:
 ```sh
 $ poetry run bia-converter-light convert-image da612702-e612-4891-b440-816d0a2b15be
 ```
 This assumes that the UUID supplied (`da612702-e612-4891-b440-816d0a2b15be`) is that of an image representation in the API
```
liviuba, line 38: Just wondering, is there an easy way to create the dependencies for one run of `convert-image`? As in, I'd like to run it once on my local on a fresh db, how can I get a (dummy) image representation to run it on?

liviuba: oh, or maybe some ingest commands on real data but known to be small / useable as a test?

bia-ingest/bia_ingest/conversion/image_annotation_dataset.py
```diff
@@ -14,13 +13,16 @@
...
 
 
 logger = logging.getLogger("__main__." + __name__)
 
 
 def get_image_annotation_dataset(
-    submission: Submission, result_summary: dict, persist_artefacts=False
+    submission: Submission,
+    result_summary: dict,
+    persister: Optional[PersistenceStrategy | None] = None,
```
liviuba, line 25: I think Optional[Thing] is defined as `Thing | None`, so the one in brackets is redundant (here and in a few other places)

bia-converter-light/README.md
liviuba: nvm just realised some small changes for bia-converter-light got bundled in but the PR is mainly for ingest (dummy data to run on local would still be useful IMO, but out of scope)

kbab:
> Should the default be on disk? I'm a bit worried about the default making changing to production data...

I am happy to change this. My impression was that the most common use would be for the API, but I guess users may want to explore things locally before running against the API. @liviuba do you have a preference for the default?

liviuba:
> > Should the default be on disk? I'm a bit worried about the default making changing to production data...
> 
> I am happy to change this. My impression was that the most common use would be for the API, but I guess users may want to explore things locally before running against the API. @liviuba do you have a preference for the default?

No strong preference aside from really wanting 'Undo' functionality in the api :) @sherwoodf what's the scenario you had in mind where data would be accidentally written to prod?

sherwoodf:
This morning i ran it assuming it would write locally (because that how it currently works) and would have submitted data somewhere were it not for the fact that i got some error (see above).

I get that in most cases we do want to send data to the api, but without a magic undo button it feels there should be an intentional flag to say 'please actually do this' when we may also want to run the (otherwise) exact same command for other reasons.

liviuba:
> This morning i ran it assuming it would write locally (because that how it currently works) and would have submitted data somewhere (i don't know where, i wasn't paying attention to this) were it not for the fact that i got some error (see above).
> 
> I get that in most cases we do want to send data to the api, but without a magic undo button it feels there should be an intentional flag to say 'please actually do this' when we may also want to run the (otherwise) exact same command for other reasons.

To me, [this](https://github.com/BioImage-Archive/bia-integrator/pull/214#issuecomment-2390956083) looks like a bug for a missing validation where the tools fails in an unexpected way if it's missing config items for what it's executed for

liviuba (changes requested):
Few suggestions

* The `Optional[Thing | None]`
* The bug @sherwoodf found, a validation somewhere that picks up if the app is missing configs for the mode it's called for would be nice IMO
* (happy to merge without, but mentioning in the spirit of tests being complete / green ticks meaning functional) Is it feasible to make the tests cover the code path where the api is involved?

bia-ingest/bia_ingest/conversion/image_annotation_dataset.py
kbab: | is from 3.10+. Since we are requiring this in pyproject.toml I will make the change across the board. Thanks

liviuba: Just to clarify, I meant that `Optional[PersistenceStrategy | None] == PersistenceStrategy | None | None` (IIRC), so the `| None` in the brackets is redundant, this wasn't about `|` itself

kbab:
> > > Should the default be on disk? I'm a bit worried about the default making changing to production data...
> > 
> > 
> > I am happy to change this. My impression was that the most common use would be for the API, but I guess users may want to explore things locally before running against the API. @liviuba do you have a preference for the default?
> 
> No strong preference aside from really wanting 'Undo' functionality in the api :) @sherwoodf what's the scenario you had in mind where data would be accidentally written to prod?

I have changed default to disk.

kbab:
> Minor comment, but i think it would be better if the 'written' log lines were debug level - these can always be viewed with a -v, and would reduce the quantity of output when dealing with long file lists. We might want to revisit some of the debug logs as well. 

Changed the `INFO` to `DEBUG`.

bia-ingest/test/test_bia_ingest_cli.py
```diff
@@ -52,7 +53,15 @@ def _load_submission(accession_id: str) -> biostudies.Submission:
 
     monkeypatch.setattr(cli, "load_submission", _load_submission)
 
-    result = runner.invoke(cli.app, ["ingest", accession_id])
+    result = runner.invoke(
```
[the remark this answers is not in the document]
kbab: Yes, it would be ideal to test - I have created [this clickup ticket](https://app.clickup.com/t/8696229nr). I will discuss implementation with you. The issue I have is that if I want to test with api locally, I would want a *clean* instance of the local api each time the test is run. But I also use my local api for dev purposes. I will explore getting pytest to spin up its own api and mongo instance.

kbab:
> > This morning i ran it assuming it would write locally (because that how it currently works) and would have submitted data somewhere (i don't know where, i wasn't paying attention to this) were it not for the fact that i got some error (see above).
> > I get that in most cases we do want to send data to the api, but without a magic undo button it feels there should be an intentional flag to say 'please actually do this' when we may also want to run the (otherwise) exact same command for other reasons.
> 
> To me, [this](https://github.com/BioImage-Archive/bia-integrator/pull/214#issuecomment-2390956083) looks like a bug for a missing validation where the tools fails in an unexpected way if it's missing config items for what it's executed for
> 
> LE: Happy for the default to be anything, but it would fail in the same way if it had a `--save-to-api` flag but no config set

Added assertions in __init__ methods to check valid parameters are supplied.

bia-ingest/bia_ingest/persistence_strategy.py
```diff
@@ -72,6 +78,9 @@ def fetch_by_uuid(
 # Persist using API
 class ApiPersister(PersistenceStrategy):
     def __init__(self, api_client: PrivateApi) -> None:
+        assert isinstance(
+            api_client, PrivateApi
+        ), f"ApiPersister cannot be created. Expected valid instance of <class 'PrivateApi'>. Got : {type(api_client)} - are your API credentials valid and/or is the API server online?"
```
liviuba, line 83: +1 "how to start fixing the error" in the error always useful!

liviuba (approved):
LGTM!

PR #216, opened 2024-10-04
added check for count of files and dryrun mode

sherwoodf:
https://app.clickup.com/t/8695y91m7

Added check using the biostudies info tables before trying to process filelists

check has a y/N question for any studies with more than 200k files.

Example command to test:

poetry run biaingest ingest  S-BIAD1069 --process-filelist=skip --dryrun

### DIFF
--- a/bia-ingest/bia_ingest/cli.py
+++ b/bia-ingest/bia_ingest/cli.py
@@ -204,7 +204,6 @@ def create(
 def determine_file_processing(
     process_files_mode: ProcessFilelistMode, file_count_limit: int, file_count: int
 ) -> bool:
-    file_count_limit = 200000
     if process_files_mode == ProcessFilelistMode.always:
         process_files = True
     elif process_files_mode == ProcessFilelistMode.skip:
--- a/bia-ingest/bia_ingest/cli.py
+++ b/bia-ingest/bia_ingest/cli.py
@@ -1,7 +1,8 @@
 import typer
 from typing import List
+from enum import Enum
 from typing_extensions import Annotated
-from bia_ingest.biostudies import load_submission
+from bia_ingest.biostudies import load_submission, load_submission_table_info
 from bia_ingest.config import settings, api_client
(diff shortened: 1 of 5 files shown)


bia-ingest/bia_ingest/cli.py
```diff
@@ -128,12 +155,14 @@ def create(
 
     result_summary = {}
 
-    persister = persistence_strategy_factory(
-        persistence_mode,
-        output_dir_base=settings.bia_data_dir,
-        accession_id=accession_id,
-        api_client=api_client,
-    )
+    persister = None
```
kbab, line 158: This causes an issue for creating image representations - it goes back to your question about having separate readers and writers. As part of the creation of an image representation, the file reference needs to be read.

My approach to fixing this would be to have  a flag on the persister that if set does not persist, and logs that it is not persisting because it is a dry run. This will also be informative for the user. I am happy to implement this now and commit it to this branch.

sherwoodf: Got it - will revisit the representation code next week to understand it better.

bia-ingest/bia_ingest/cli.py
```diff
@@ -174,6 +201,23 @@ def create(
     print(f"\n\n[red]--------- Errors ---------\n{errors}[/red]")
 
 
+def determine_file_processing(
+    process_files_mode: ProcessFilelistMode, file_count_limit: int, file_count: int
+) -> bool:
+    file_count_limit = 200000
```
kbab, line 207: `file_count_limit` is also assigned a value in line 105. Is it worth defining a global constant for this so value set in only one place?

sherwoodf: Oh oops. I should just remove it from here.

kbab (commented):
LGTM and happy to merge as-is. Just one comment about possibly using a constant. Please re-request review either way.

kbab (approved):
LGTM

PR #218, opened 2024-10-07
Ingest refactor

sherwoodf:
Tried to be clear with each  commit and make them individual pieces of work. Overall tried to refactor under the following guidance from discussions with kola and liviu:

1. From Kola: The representation creation part of the code starts with BIA objects, and so should probably be able to be disconneced from all biostudies code
2.  From Liviu: All functions are something's 'utils' functions, so i tried to be more explicit with file naming (even if a few utils files remain) in order to provide a better idea of what kind of functions should exist there.

I would say this doesn't touch any business logic. The closest I come to are some changes (simplifications) to logging in the representation creation cli because you can only use it on 1 study at the same time, so there is no need to pass around a) a dictionary of results (keyed by study Accession ID) and b) the submission (which was just being used to provide the right key for the results dictionary). The only other near business logic change is switching a deprecated pydantic class parameter to it's new counterpart. The result of the changes are that the code in representation_creation folder does not import anything from ingest and vice-versa.

I also made some updates to the readme to try to be more explicit about what the package does & expects.

The changes to the ingest code were only to move functions around. I created a biostudies folder to house all the code we use to read biostudies entries - with the idea that we would later create more folders for different ingestion locations. But i suspect the various convert code makes too many assumptions that it's ingesting from biostudies, and we will probably end up with multiple folders along the lines of: ingest_biostudies, ingest_empiar etc.

I also deleted all the test data directories with example jsons. They're out of date, and we now have actual jsons we generate to compare to.

I'd like to do a 2nd pass after this to clarify that experimentally captured images are created with the representation command.


bia-ingest/README.md
```diff
@@ -1,57 +1,95 @@
...
         - bia_api_basepath
         - bia_api_username
         - bia_api_password
-    * Alternatively, to read/persist to disk the default is `~/.cache/bia-integrator-data-sm/` which can be changed by setting `bia_data_dir`
+    * When reading/persisting to disk the default location is `~/.cache/bia-integrator-data-sm/`. This can be changed by setting `bia_data_dir`.
+
+## Usage
+This package has 2 cli applications:
+ * ingest: used to transform BioStudies submissions into various BioImage Archive API objects
+ * represetations: used to create Image representations from BioImage Archive File Reference objects.
```
kbab, line 14: Minor comment: represetations -> representations

bia-ingest/README.md
```diff
@@ -1,57 +1,95 @@
...
 
-To ingest into the api, the persistence mode and one or more accession ids have to be supplied. Assuming you are in this directory:
+Note that this package does not convert images (i.e. create image files that contain pixel data) from existing objects.
+This is now handled by the [bia-converter-light](../bia-converter-light/README.md) sub package.
+This package only creates the objects that exist in the BIA API, which is used to store metadata about such images.
+
+## Ingest Commands
+To create BIA API objects from one or more biostudies submissions, assuming the package was installed via peoetry and you are running from the same directory as this readme, run:
```
kbab, line 21: peoetry -> poetry

bia-ingest/README.md
```diff
@@ -1,57 +1,95 @@
...
+The STATIC_DISPLAY representation is not created by default because the BIA website only needs one static display per experimental imaging dataset. All interactive images need a thumbnail for the website, so they are usually created together.
 
 An option can be passed into the command to specify representations to create. E.g. to create only THUMBNAIL and STATIC_DISPLAY:
-```
-biaingest representations create --reps-to-create THUMBNAIL --reps-to-create STATIC_DISPLAY S-BIAD1348 a9125402-5e47-4afd-9abf-7393448acd07 e1c99f04-549d-4952-ba88-c921c707f01d f26e27e4-90ce-42b4-91ae-503dafed8b70
+```sh
+$ poetry run biaingest representations create --reps-to-create THUMBNAIL --reps-to-create STATIC_DISPLAY S-BIAD1285 002e89fc-5a6c-4037-86ec-0dadd9553694
 ```
 
 ## Converting images
```
kbab, line 95: There is no content under this heading. Should it be deleted?

bia-ingest/bia_ingest/bia_object_creation_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    api_models = []
+    for model_dict in dicts:
+        try:
+            api_models.append(api_model_class.model_validate(model_dict))
+        except ValidationError:
+            log_failed_model_creation(api_model_class, valdiation_error_tracking)
+    return api_models
+
+
+def get_bia_data_model_by_uuid(
```
kbab, line 47: I don't think we need this anymore - it was useful before we had the API/persistence to disk, but the persister can perform this functionality now. As far as I can tell it is not used in any current modules.

bia-ingest/bia_ingest/cli_logging.py
```diff
@@ -1,9 +1,18 @@
+from typing import Type
 from rich.table import Table
 from rich.text import Text
 from pydantic import BaseModel, Field
+import logging
+from abc import ABC
```
kbab, line 6: I can't see where this is used...

bia-ingest/bia_ingest/cli_logging.py
```diff
@@ -46,13 +55,16 @@ class IngestionResult(BaseModel):
     Contributor_ValidationErrorCount: int = Field(default=0)
     Organisation_CreationCount: int = Field(default=0)
     Organisation_ValidationErrorCount: int = Field(default=0)
+
+
+class ImageCreationResult(CLIResult):
```
kbab, line 60: +1

kbab (approved):
LGTM

PR #220, opened 2024-10-09
Pylance complaining about variables used in type hints

sherwoodf:
My Pylance currently complains:

Variable not allowed in type expressionPylance[reportInvalidTypeForm](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportInvalidTypeForm)

I don't know that we want to make this change, but i checked that the openapi.json generated by this and the one currently generated by main are the same (neither are the same as the one currently in the client). In any case this was useful to me to better understand how the api works.

### DIFF
--- a/api/api/public.py
+++ b/api/api/public.py
@@ -36,8 +36,12 @@ def make_get_item(t: Type[shared_data_models.DocumentMixin]):
     # @TODO: nicer wrapper?
     async def get_item(
         uuid: shared_data_models.UUID, db: Annotated[Repository, Depends(get_db)]
-    ) -> t:
+    ) -> Type[shared_data_models.DocumentMixin]:
         return await db.get_doc(uuid, t)
+
+    # Setting the annotation for the client to generate correctly.
+    # Variable type annotations are currently not allowed in python typing spec, so setting it manually after creating the function.
+    get_item.__annotations__["return"] = t
 
     return get_item
 
@@ -49,7 +53,7 @@ def make_reverse_link_handler(
 ):
     async def get_descendents(
         uuid: shared_data_models.UUID, db: Annotated[Repository, Depends(get_db)]
-    ) -> List[source_type]:
+    ) -> List[Type[shared_data_models.DocumentMixin]]:
         # Check target document actually exists (and is typed correctly - esp. important for union-typed links)
         #   see workaround_union_reference_types
         await db.get_doc(uuid, target_type)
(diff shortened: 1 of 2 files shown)


api/api/public.py
```diff
@@ -36,9 +36,13 @@ def make_get_item(t: Type[shared_data_models.DocumentMixin]):
     # @TODO: nicer wrapper?
     async def get_item(
         uuid: shared_data_models.UUID, db: Annotated[Repository, Depends(get_db)]
-    ) -> t:
+    ) -> Type[shared_data_models.DocumentMixin]:
```
liviuba, line 39: IMO this makes the code less obvious because typing is expected to be "true" in fastapi apps.

Strictly speaking, this is a code workaround for an IDE warning, where we set the return type to an inaccurate-but-not-confusing-for-pylance one and then set it to the correct one below, with the same results as before but just not detected by pylance.

I'd rather fix it on the pylance side by adding a "don't check this" annotation, or in the config and keep the code as clear as possible. I would guess it's pylance being strict rather than variable return annotations being undefined behaviour / explicitly forbidden (definitely not because it works) in python

Thougts?

sherwoodf: Just to be clear, i don't like this particular solution either. 

Practically, i'm fine with closing this PR and figuring out what setting i should change in VS code to not get the error. 

I suspect the ideal solution would be to change something in the code in a way that's better than this.

From a 'python standards' point of view, i'm getting the impression that this is supposed to be explicitly forbidden, it's just not been made explicit quickly enough, and a bunch of libraries are now using annotations in a way the type annotation people aren't expecting/don't want them to. My tl;dr version of the issues i've seen would be:

The Type annotations are designed to provide type hints in IDEs. The standard is designed to provide gradual, static typing to python. Therefore any type hints which require code to be run in order to work out what the type should be are disallowed. The people in charge of the type spec have been improving test coverage, and an update in February of this year means this and a bunch of other things now show as errors.

It seems like blender is using type hints in a way in which the spec doesn't expect, and there's lots of 'broken workflows': https://github.com/microsoft/pylance-release/issues/5457. I could imagine FastAPI having done something similar.

In any case i feel like digging into this is helping me better understand the API code, fast api, typing, etc so i don't mind spending a bit more looking into if there might be a better solution.

liviuba: PR with a redesign for something better is welcome! Not tested, but the pylance fix seems to be

```json
```

PR #222, opened 2024-10-16
major model update

sherwoodf:
Note this will require a lot of changes in the other packages & a data migration.

Notes on changes - check out picture in readme (or here https://docs.google.com/drawings/d/1y_0GjhIxhB1hVvjIBoTyEv8_SE1jXuJ_9-QK6ScouBg/edit) for overview

Created new CreationProcess object - this has fields for each class of object, even though some objects play similar semantic role. Another option would have been to have specific subclasses for each type of creation process, but then the image -> creation process link would have been more complex.

Added a AnnotationSourceIndicator to annotation methods to try to capture e.g. file name convention for annotation files (might make automated processing easier)

Move around the classes in mock objects & bia_data_models so that they all follow the same pattern.

Clickup tickets:

https://app.clickup.com/t/8696348kx
https://app.clickup.com/t/86962c8y3
https://app.clickup.com/t/86968w0ba


bia-shared-datamodels/pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 packages = [{include = "bia_shared_datamodels", from = "src"}]
 
 [tool.poetry.dependencies]
-python = "^3.10"
+python = "^3.10,<3.12"
```
kbab, line 11: Is there a reason for restricting to <3.12?

bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
```diff
@@ -146,12 +146,32 @@ class UserIdentifiedObject(BaseModel):
...
+#######################################################################################################
+
+
 class Study(
     semantic_models.Study,
     DocumentMixin,
 ):
     author: List[semantic_models.Contributor] = Field(min_length=1)
 
+    model_config = ConfigDict(model_version_latest=2)
```
kbab, line 160: Did we make a decision about re-ingesting all data or migrating? If we are re-ingesting all data and will have no record of previous versions, do we need to increment the version number?

bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
```diff
@@ -164,94 +184,109 @@ class FileReference(
...
     representation_of_uuid: Annotated[
         UUID,
-        ObjectReference(
-            None,
-            workaround_union_reference_types=[
-                DerivedImage,
-                ExperimentallyCapturedImage,
-            ],
-        ),
+        ObjectReference(None, Image),
```
kbab, line 220: Should second parameter be `[Image]` ?

sherwoodf: Yes, this was because i started off with the union, and then made lots of individual fields, but just moved the types out. Will fix

bia-shared-datamodels/pyproject.toml
sherwoodf: Mostly just to keep consistency between this and our other python packages  - which avoid 3.12 because of the " vs ' for fstring thing that really confused me for a while.

bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
sherwoodf: I think that in either case, we would have no record of previous version - it's more that if something went wrong with either the migration or i guess the dropping of the database (though that's less likely) and we had some data hanging around, we might be able to track it down.

liviuba (approved):
LGTM!

PR #223, opened 2024-10-16
Add single settings class for api settings

liviuba:
https://app.clickup.com/t/8695q6qjd

Most changes were workarounds to cyclical imports from the get_db dependency being moved to app.py


api/api/auth.py
```diff
@@ -1,23 +1,24 @@
...
+from api.models.repository import Repository
 from api.models.persistence import User
 from api.models.api import AuthenticationToken, TokenData, AuthResult
 import os
 import base64
 
+
 router = APIRouter(prefix="/auth", tags=[constants.OPENAPI_TAG_PRIVATE])
 
 JWT_SECRET_KEY = os.environ["JWT_SECRET_KEY"]
```
kbab, line 24: Should this be using settings?

api/api/auth.py
```diff
@@ -1,23 +1,24 @@
...
-from typing import Optional
 import uuid
 
+from api.app import get_db
 from api import constants
-from api.models.repository import Repository, get_db
+from api.models.repository import Repository
 from api.models.persistence import User
 from api.models.api import AuthenticationToken, TokenData, AuthResult
 import os
```
kbab, line 18: Is this needed anymore?

kbab (changes requested):
I left some comments on auth.py. Also in conftest.py:98 need to use settings and do not need `import os`.

A suggestion - which is not a requirement for approval is to make the initialisation on line 18 of `settings.py` absolute (e.g. `env_path = Path(__file__).parent / ".env" ` which will also simplify use of package. E.g. for testing I would not need to be in `api/api`. Happy to discuss after lunch.

Finally, not sure if it is my local config, but two tests for auth failed:
```
=========================== short test summary info ============================
FAILED tests/test_auth.py::test_create_user - assert 500 == 200
FAILED tests/test_auth.py::test_create_user_bad_token - AssertionError: ['Traceback (most recent call last):
=================== 2 failed, 36 passed, 1 warning in 10.44s ===================
```

api/api/auth.py
liviuba: yes! Missed it

kbab (changes requested):
LGTM - just one change request and one comment

1. Change: api/api/tests/conftest.py:101 tries to access env variable.
2. I was wondering why asserts are used in the fixtures in conftest - should these be tests in their own right?

liviuba:
Fixed 1., missed it!

For 2, the asserts are there just to make sure nothing went wrong. They're not meant to be tests, just to build dependencies for tests, but it's nice to check the response for things.

kbab (approved):
LGTM!

PR #224, opened 2024-10-17
update settings

sherwoodf:
Update vscode settings to find the client in the correct location

For some reason my vscode cannot find the client - it is always installed correctly, but i think it's in too deep a folder and vscode just gives up before finding it.


liviuba (approved):
LGTM! 

"python.defaultInterpreterPath" in every sub-project also fixes annoying IDE warnings when I have the main `bia-integrator` directory open! Very useful

PR #225, opened 2024-10-18
Client gen in docker compose

liviuba:
https://app.clickup.com/t/869643en7


clients/Readme.md
```diff
@@ -1,31 +1,5 @@
...
-```
-
-
-docker run -it --rm \
-    -v /etc/passwd:/etc/passwd:ro -v /etc/group:/etc/group:ro --user $(id -u) \
-    -v ${PWD}:/mnt/share \
-    -w /mnt/share \
-    openapitools/openapi-generator-cli:v7.8.0 \
-    author template -g python -o python_template
+make client.generate
```
kbab, line 4: PR looks good to me and works on my local - its good to have this! Wanted to suggest specifying that this make command refers to Makefile in directory above. Or use command `make -C .. client.generate` . However, happy to approve as is. Just let me know what you think.

kbab (approved):
LGTM

clients/Readme.md
liviuba: Agree, added the change, thanks!

(tests failing because of a random dependency now, really should add `poetry.lock` back :D, but shout fix themselves)

PR #227, opened 2024-10-22
Fix api image to match deployment, add new env vars to helm chart

liviuba:
Half of https://app.clickup.com/t/86964ygd3


sherwoodf (approved):
LGTM

PR #230, opened 2024-10-23
Make accessing values of dicts created from sections during ingest insensitive to case of keys

kbab:
Fix to bug arising from difference in cases of names of some sections in submission tool forms and BIA yaml template. E.g.
`Sample preparation protocol` in [S-BIAD1327](https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1327) submitted using submission tool and `Sample Preparation Protocol` in [S-BIAD1117](https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1117) submitted using yaml template.

This fix creates a function to access dictionaries using case insensitive keys and is applied to all instances in ingest where mapping occurs using names from biostudies sections.


sherwoodf (approved):
LGTM - i never know whether we would want to add tests for this kind of thing in the future, but i'm happy to merge without for now.

PR #231, opened 2024-10-24
update tests to fit new models

sherwoodf:
Will require a rebase/rerun of the tests after: https://github.com/BioImage-Archive/bia-integrator/tree/refs/heads/model_updates_1 is merged, but this should get all the api tests working.


api/api/tests/test_union_types.py
```diff
@@ -0,0 +1,55 @@
+# Note: This file is all commented out because the feature, which enabled us
```
liviuba, line 1: If it's unlikely we'll have union links (so could use this as a quick template), I'd rather delete this file altogether for tidyness

liviuba (approved):
LGTM! Just a minor comment for tidyness

api/api/tests/test_union_types.py
sherwoodf: It does seem unlikely for now, so i'm happy to delete (i'd still like to keep the feature i guess, as that would be a bit more of a faf to re-create)

PR #235, opened 2024-10-29
Uncomment uuid uniqueness test, fix fixture naming

liviuba:
https://app.clickup.com/t/8696c19qp

No edge cases in the new models I could find. Small rename to match model name exactly & uncomment a test

And https://app.clickup.com/t/8696c2dn2 - this was added because looking at the models and the endpoint names in http://localhost:8080/redoc was confusing. Just reversed the attributes and added a description


kbab (approved):
Looks good to me. I noted /src/models/repository.py:18 still uses env vars. Is this OK?

PR #236, opened 2024-10-29
Setup client examples to run in ci

liviuba:
https://app.clickup.com/t/8696c1bc2

Needs https://github.com/BioImage-Archive/bia-integrator/pull/235

### DIFF
--- a/clients/python/example/private_api.py
+++ b/clients/python/example/private_api.py
@@ -3,12 +3,22 @@ Example script that creates and gets some objects. Please see the api client rea
 """
 
 from bia_integrator_api.util import get_client_private
-from bia_integrator_api.models import Study, LicenceType, Contributor, ExperimentalImagingDataset
+from bia_integrator_api.models import Study, LicenceType, Contributor, Dataset
 from datetime import date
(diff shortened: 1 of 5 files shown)


clients/python/example/private_api.py
```diff
@@ -17,7 +27,7 @@
 
 # Example create/get - equivalent for every object
 study = Study(
-    uuid="06c19696-00e8-4c2e-a27f-23587aedb782",
+    uuid=get_uuid(),
     version=0,
     accession_id="test",
```
kbab, line 32: I had to change this to S-BIADTEST to get things working without errors on my local ...

liviuba: will change the search to be for something that's actually created before, thanks! It was finding in the db studies created for the api tests, in https://github.com/BioImage-Archive/bia-integrator/blob/9a72caaa27a281a6094e417745a3c6fc04d00292/bia-shared-datamodels/src/bia_shared_datamodels/mock_objects.py#L25

kbab (approved):
LGTM!

PR #237, opened 2024-11-05
Api proxy root

liviuba:
https://app.clickup.com/t/8695pw50c


liviuba:
> LGTM. Only comment is I could not find any info on significance of fastapi_root_path. Would adding it to .env_template with a comment be useful to users?

Oh yes, should've been in the template, fixing now

kbab (approved):
LGTM

PR #239, opened 2024-11-11
Refactor tests - create mock objects in their own modules

kbab:
bia-ingest/test/utils.py became very long (> 900 lines). This PR refactored by creating a mock_objects sub-package with modules for each mock object - [https://app.clickup.com/t/8695yd3r0](https://app.clickup.com/t/8695yd3r0).


bia-ingest/test/mock_objects/utils.py
```diff
@@ -0,0 +1,131 @@
+"""Utility functions for mock objects"""
+
+accession_id = "S-BIADTEST"
+
+# TODO: Move commented out functions for imaging to bia-converter-light
```
sherwoodf, line 5: Somewhat surprised to not see these moved as part of this PR rather than just commenting out (even if they're not used in the converter).

sherwoodf (approved):
Approved with just a comment about comments

PR #242, opened 2024-11-13
updated models with attributes and title_id

sherwoodf:
Updates to lots of classes to add an attribute field (done via mixin, as it seemed like this would help keep the name & design consistent in the future), and added requried title_id to Protocols


kbab (approved):
LGTM

PR #243, opened 2024-11-13
updated client after model update

sherwoodf:
Update client from new models, adding attribute field to many classes (note that in some, it has modified the order of the fields, but the document hasn't otherwise changed) and adding title_id to protocols


liviuba (approved):
LGTM!

PR #244, opened 2024-11-13
Small tweaks for better speed

liviuba:
For this https://app.clickup.com/t/8696kk2bj and this https://app.clickup.com/t/8696kmfu6

### DIFF
--- a/api/api/settings.py
+++ b/api/api/settings.py
@@ -1,10 +1,15 @@
 from pydantic_settings import BaseSettings, SettingsConfigDict
+from pydantic import Field
 from pathlib import Path
 
 
 class Settings(BaseSettings):
     mongo_connstring: str
     mongo_timeout_ms: int
+    mongo_max_pool_size: int = Field(
+        default=10,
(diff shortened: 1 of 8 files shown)


api/api/settings.py
```diff
@@ -1,10 +1,15 @@
...
+from pydantic import Field
 from pathlib import Path
 
 
 class Settings(BaseSettings):
     mongo_connstring: str
     mongo_timeout_ms: int
+    mongo_max_pool_size: int = Field(
+        default=10,
+        description="Max size of the Mongo connection pool. If 1 process/api, this should be higher if the link to Mongo is fast",
```
kbab, line 11: I am not clear what the statement "If 1 process/api," means.

liviuba: updated the text now

liviuba:
Just noticed tests fail, it's because of the on_start changes, I'll fix that later, it's ok to ignore for now

kbab (approved):
LGTM (apart from tests issue)

PR #247, opened 2024-11-19
Logging features fixes

sherwoodf:
- Fix to biosample logic so that creation of biosamples is logged regardless of persistence strategy (was causing a bug with the --dryrun option)
- Added generic error catching which get logged in the output table, as well as an option to write the table to a csv for easier analysis of results.


sherwoodf:
superseded by: https://github.com/BioImage-Archive/bia-integrator/pull/248

PR #249, opened 2024-11-28
Non-breaking model updates

sherwoodf:
added licence enum options and make Analysis and Correlation method models have optional fields

Also fix to tests failing on github due to different url types.

https://app.clickup.com/t/8696v6vyg

https://app.clickup.com/t/8696rzer5

Will update API client after this is merged.

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/mock_objects.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/mock_objects.py
@@ -6,7 +6,7 @@ from bia_shared_datamodels import semantic_models
 from uuid import uuid4
 from enum import Enum
 import datetime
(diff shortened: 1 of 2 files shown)


bia-shared-datamodels/src/bia_shared_datamodels/mock_objects.py
```diff
@@ -6,7 +6,7 @@
 from uuid import uuid4
 from enum import Enum
 import datetime
-from pydantic_core import Url
+from pydantic import AnyUrl
```
sherwoodf, line 9: Tests were failing in github, but not locally. Unsure whether something changed that caused this?

kbab (approved):
LGTM

PR #250, opened 2024-11-28
Update client

sherwoodf:
update client after licence changes and some fields have been made optional. For model changes, see: https://github.com/BioImage-Archive/bia-integrator/pull/249


liviuba (approved):
LGTM!

PR #252, opened 2024-11-29
added functions for uuid creation

sherwoodf:
https://app.clickup.com/t/8696w910y

I've left image, image representation, specimen, creation process, commented out - suggestions on which to build from, but i'm no expert there.

I've also left out annotation files - we don't yet create these, and i'm not sure what they would need (maybe just file ref uuid & type?)

I'm not sure what or how we would want to test these?


kbab (approved):
LGTM. As we discussed on 28/11/2024 using (accession_id, title, bia_data_model.__name__) triplet can result in rare instances of multiple biostudies sections with same titles being assigned same UUID. This is extremely rare - I have only been able to find one instance ([S-BIAD1040](https://www.ebi.ac.uk/biostudies/bioimages/studies/S-BIAD1040) - see biosamples with accno Biosample-1 and Biosample-6 in [study json](https://www.ebi.ac.uk/biostudies/files/S-BIAD1040/S-BIAD1040.json)) in the hundreds of submissions we have so far.

PR #254, opened 2024-12-02
deal with ontology values in subsections of protocols/biosample

sherwoodf:
Update to deal with ontology structured subsections in image acquisitions and biosamples a good study to test with is:

`poetry run biaingest ingest --process-filelist skip S-BIAD43 -v -c`

https://app.clickup.com/t/8696v7mpx

Also updated the study -> processing mode list to cover some of the bia-agent ingested studies. We could alternatively try to map on other keys in the submission.

https://app.clickup.com/t/8696v7zke

### DIFF
--- a/bia-ingest/bia_ingest/biostudies/v4/bio_sample.py
+++ b/bia-ingest/bia_ingest/biostudies/v4/bio_sample.py
@@ -7,22 +7,22 @@ from bia_ingest.biostudies.generic_conversion_utils import (
 )
 from copy import deepcopy
 
-
-from ...bia_object_creation_utils import (
+from bia_ingest.bia_object_creation_utils import (
     dict_to_uuid,
+    dicts_to_api_models,
     dict_map_to_api_models,
     filter_model_dictionary,
 )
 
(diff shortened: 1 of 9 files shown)


bia-ingest/bia_ingest/biostudies/v4/bio_sample.py
```diff
@@ -144,24 +146,47 @@ def generate_biosample_uuid(biosample_dict: dict[str, Any]) -> str:
...
+    if subsection_organisms:
+        key_mapping = [
+            ("common_name", "Common name", None),
+            ("scientific_name", "Scientific name", None),
+            ("ncbi_id", "NCBI taxon ID", None),
+        ]
+        for section in subsection_organisms:
+            attr_dict = attributes_to_dict(section.attributes)
+
+            model_dict = {
```
kbab, line 164: This re-assignment does not change the value of the passed in model_dict, but I had to look closely to see if this was the intended behaviour...

kbab (approved):
LGTM

PR #259, opened 2024-12-04
use new uuid creation functions

sherwoodf:
Update ingest to use UUID functions

Updated bio sample uuid function to use growth protocol uuid to make sure it's unique.

Remove dict_to_uuid and filter_model functions from ingest, since they are no longer needed

Changed file reference functions slightly to a) follow pattern of other objects (creating list of dicts and then converting the dicts) and b) log the total number of created filelists. This ended up meaning some changes to the filelist tests - i decided to move some files from study component 1 to the annotation component so we can test that we definitely create annotation component file references in datasets.


kbab (approved):
LGTM!

PR #260, opened 2024-12-04
scripts for analysis of studies in the BIA collection

sherwoodf:
Thought i should share the scripts i used to create the analysis of fields that appear in studies in biostudies

script also create a list of all the biostudies studies we want to ingest, so it's useful to run them to get an up-to date list of studies.

### DIFF
--- a/bia-ingest/scripts/biostudies_fields_stats.py
+++ b/bia-ingest/scripts/biostudies_fields_stats.py
@@ -1,0 +1,138 @@
+import logging
+from rich import print
+from rich.logging import RichHandler
+from bia_ingest.biostudies.api import load_submission, Section
+from pydantic import BaseModel, ValidationError
+from typing import List, Optional
+import json
+import csv
+import time
+
+logging.basicConfig(
+    level=logging.INFO, format="%(message)s", handlers=[RichHandler(show_time=False)]
+)
+logger = logging.getLogger()
+
+
+class SearchResult(BaseModel):
+    accession: str
+    type: str
+    title: str
+    author: str
+    links: int
+    files: int
+    release_date: str
+    views: int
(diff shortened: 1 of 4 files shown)


bia-ingest/scripts/biostudies_fields_stats.py
```diff
@@ -0,0 +1,138 @@
...
+    totalHits: int
+    isTotalHitsExact: bool
+    sortBy: str
+    sortOrder: str
+    hits: List[SearchResult]
+    query: Optional[str]
+    facets: Optional[str]
+
+
+def count_section_type(dict_to_edit: dict, section_to_count: str):
```
Cesarec88, line 41: I'd suggest to simplify this code using `defaultdict`

https://docs.python.org/3/library/collections.html#defaultdict-objects

bia-ingest/scripts/biostudies_fields_stats.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+def recurse_subsection(section: Section, dict_count: dict):
+    count_section_type(dict_count, section.type)
+    if len(section.subsections) > 0:
+        for section in section.subsections:
+            if type(section) is list:
+                pass
+            else:
+                recurse_subsection(section, dict_count)
```
Cesarec88, line 54: this can be simplified with using the logical negation of the statement 
```

if type(section) is no list:
  recurse_subsection(section, dict_count)
```

bia-ingest/scripts/biostudies_fields_stats.py
Cesarec88: it is an usefull method that defaults counters of dictionaries as we define a key

PR #261, opened 2024-12-05
updated export to use new association structure in attributes

sherwoodf:
We now store the all the relevant uuids in the dataset in order to create images, so using that instead of trying to re-use the associations that are also stored as-is in attributes.

### DIFF
--- a/bia-export/test/input_data/dataset/S-BIADTEST/47a4ab60-c76d-4424-bfaa-c2a024de720c.json
+++ b/bia-export/test/input_data/dataset/S-BIADTEST/47a4ab60-c76d-4424-bfaa-c2a024de720c.json
@@ -4,6 +4,27 @@
     "description": "Description of study component 1",
     "attribute": [
         {
+            "provenance": "bia_ingest",
+            "name": "image_acquisition_protocol_uuid",
+            "value": {
+                "image_acquisition_protocol_uuid": ["c2e44a1b-a43c-476e-8ddf-8587f4c955b3"]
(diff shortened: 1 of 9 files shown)


bia-export/test/input_data/dataset/S-BIADTEST/47a4ab60-c76d-4424-bfaa-c2a024de720c.json
```diff
@@ -3,6 +3,27 @@
...
+            "name": "specimen_imaging_preparation_protocol_uuid",
+            "value": {
+                "specimen_imaging_preparation_protocol_uuid": ["7199d730-29f1-4ad8-b599-e9089cbb2d7b"]
+            }
+        },
+        {
+            "provenance": "bia_ingest",
+            "name": "bio_sample_uuid",
+            "value": {
+                "bio_sample_uuid": ["64a67727-4e7c-469a-91c4-6219ae072e99", "6950718c-4917-47a1-a807-11b874e80a23"]
```
AybukeKY, line 24: Is it bio_sample_uuid not biosample_uuid due to to_snake converting BioSample into bio_sample due to title case?

AybukeKY: was it always bio_sample_uuid or has it changed cause we're using to_snake now? and if so is it compatible with the ingest and api?

PR #266, opened 2024-12-10
fix specimen and biosample export for images

sherwoodf:
Image export wasn't creating specimens, and biosample growth protocols weren't being filled in


kbab (approved):
LGTM

PR #267, opened 2024-12-10
Email validator

Chr1st0p43rR:
Added code to catch invalid email addresses, and to record what's wrong with them in the output table. 

If an email address is found to be invalid, then it is made equal to None, and the contributor will still be recorded.

Where there are multiple errors of the same type, for any given study, then only one instance of this is recorded in the output table.

### DIFF
--- a/bia-ingest/bia_ingest/biostudies/v4/study.py
+++ b/bia-ingest/bia_ingest/biostudies/v4/study.py
@@ -323,6 +323,7 @@ def get_contributor(
     author_sections = find_sections_recursive(submission.section, ["author"])
 
     contributor_dicts = []
+    email_warnings = []
     for section in author_sections:
 
         attributes = sanitise_affiliation_attribute(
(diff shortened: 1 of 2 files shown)


bia-ingest/bia_ingest/cli_logging.py
```diff
@@ -113,20 +114,36 @@ def tabulate_ingestion_errors(
...
+        if result.Warning:
+            warning_message = result.Warning
+        
+        if (error_message == "") & (warning_message == ""):
             status = Text("Success")
             status.stylize("green")
-        else:
+        elif (error_message == "") & (warning_message != ""):
+            status = Text("Success with warnings")
+            status.stylize("orange")
```
sherwoodf, line 127: nit pick - "orange"  isn't a colour XD
https://rich.readthedocs.io/en/stable/appendix/colors.html

"yellow" would probably would work for our purposes

bia-ingest/bia_ingest/biostudies/v4/study.py
```diff
@@ -378,3 +383,31 @@ def sanitise_affiliation_attribute(
...
+        accno: str
+):
+    if email_warnings != []:
+        unique_warnings = list(set(email_warnings))
+        for warning in unique_warnings:
+            result_summary[accno].__setattr__(
+                "Warning",
+                "Email warning: " + str(warning) + "\n"
+            )
+            logger.warning(f"Email warning: {warning}")
```
sherwoodf, line 413: Nit pick: maybe we want to say "Skipped invalid author email: {warning}" if we want to make it clear what the code is doing when creating the output?

sherwoodf (approved):
2 nit pick-y comments, but looks good! Approved - but let me know if it needs a re-review.

bia-ingest/bia_ingest/biostudies/v4/study.py
Chr1st0p43rR: Yes, good point. Changed it.

bia-ingest/bia_ingest/cli_logging.py
Chr1st0p43rR: I swear I saw orange... Yellow now.

sherwoodf (approved):

PR #269, opened 2024-12-12
added file-based ingest

sherwoodf:
PR for file based ingest

run with -f <path to file> instead of accession id list

TODO: should update readme


kbab (approved):
The comments I would have left are already noted in the PR description so approving

PR #272, opened 2024-12-20
Nextflow scripts for export

kbab:
* Fix minor bug due to missing page_size parameter in using api_client
* Create a nextflow script to run export of study, dataset and image metadata one accession ID at a time and an associated python script to merge the resulting jsons into one. See [https://app.clickup.com/t/86975pvwc](https://app.clickup.com/t/86975pvwc).

### DIFF
--- a/bia-export/bia_export/website_export/images/retrieve.py
+++ b/bia-export/bia_export/website_export/images/retrieve.py
@@ -12,11 +12,12 @@ import logging
 
 logger = logging.getLogger("__main__." + __name__)
 
+DEFAULT_PAGE_SIZE = 500
+
 
 def retrieve_images(
     context: ImageCLIContext,
 ) -> list[api_models.Image]:
-
     if context.root_directory:
         api_images: List[api_models.Image] = read_all_json(
             object_type=api_models.Image,
@@ -25,12 +26,14 @@ def retrieve_images(
 
     else:
         dataset_list: List[api_models.Dataset] = api_client.get_dataset_linking_study(
-            str(context.study_uuid)
+            str(context.study_uuid),
+            page_size=DEFAULT_PAGE_SIZE,
         )
         api_images = []
         for dataset in dataset_list:
             api_images += get_all_api_results(
-                dataset.uuid, api_client.get_image_linking_dataset
+                dataset.uuid,
+                api_client.get_image_linking_dataset,
             )
 
     return api_images
--- a/bia-export/nextflow/README.md
+++ b/bia-export/nextflow/README.md
@@ -1,0 +1,23 @@
+## Description
+This sub-package allows running bia-export for a set of accession ids one accession id at a time using nextflow. It can therefore be run locally if nextflow is installed or on a slurm cluster.
+
+## Setup
+
+1. Install nextflow if not present - see [https://www.nextflow.io/docs/latest/install.html](https://www.nextflow.io/docs/latest/install.html) or if on a slurm cluster try `module load nextflow`
+2. Navigate to your working directory and:
+    1. Copy `set_local_env.sh_template` -> `./set_local_env.sh` and modify as necessary. Ensure this file has no `rwx` for `other`
+    2. Copy `nextflow.config_template` -> `./nextflow.config` and comment out the `slurm` block if running locally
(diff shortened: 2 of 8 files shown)


bia-export/bia_export/website_export/images/retrieve.py
```diff
@@ -25,12 +26,14 @@ def retrieve_images(
 
     else:
         dataset_list: List[api_models.Dataset] = api_client.get_dataset_linking_study(
-            str(context.study_uuid)
+            str(context.study_uuid),
+            page_size=DEFAULT_PAGE_SIZE,
```
AybukeKY, line 30: This is assuming the number of datasets in a study won't be larger than default_page_size parameter. Works for 500 at the moment, but it'd be good to make sure all datasets are retrieved.

bia-export/bia_export/website_export/images/retrieve.py
```diff
@@ -25,12 +26,14 @@ def retrieve_images(
...
         )
         api_images = []
         for dataset in dataset_list:
             api_images += get_all_api_results(
-                dataset.uuid, api_client.get_image_linking_dataset
+                dataset.uuid,
+                api_client.get_image_linking_dataset,
```
AybukeKY, line 36: I think this also needs to have a page size parameter? If so it'd need to loop or set the page size such that all images of the dataset are retrieved.

kbab: In this case get_all_api_results is a wrapper function that calls the specified api client function. It has a default page size setting (of 20) and polls until all data received

bia-export/bia_export/website_export/images/retrieve.py
kbab: Yes - the above is a quick fix. Default can be increased if necessary

bia-export/nextflow/set_local_env.sh_template
```diff
@@ -0,0 +1,30 @@
+# Configuration for Nextflow running bia-export
+
+# For bia-integrator-api
+export BIA_API_BASEPATH="https://wwwdev.ebi.ac.uk/bioimage-archive/api"
+export BIA_API_USERNAME=''
+export BIA_API_PASSWORD=''
+export BIA_USERNAME=''
```
AybukeKY, line 7: What is BIA_USERNAME and password used for as there's already a BIA_API_USERNAME?

kbab: This is a carry over of change in naming convention. Only one version is needed - will resolve which one and deleted unneeded.

kbab: Looked at the code and actually they are not required as we are only reading using the public api. Will delete lines 4-7!

bia-export/nextflow/README.md
```diff
@@ -0,0 +1,23 @@
...
+1. Navigate to your working directory (where `nextflow.config` was copied to)
+1. Create a manifest of the form study_uuid:accession_id and save as `./tmp/manifest.txt`. E.g.
+```
+00357b15-5a36-48e3-ab9d-9af5b697afef:S-BSST445
+0077e2c7-567a-46fa-8848-f39a86894755:S-BIAD851
+0086e8f0-8c99-4fd2-8971-833c147443c8:S-BIAD821
+00da4466-84cd-484e-adf5-2520c9c8ff0c:S-BIAD465
+```
+2. Set up the local environment: `source ./set_local_env.sh`
+3. Run export: `nextflow run $BIA_EXPORT_DIR/nextflow/run_bia_export.nf`
```
AybukeKY, line 22: I got "WARN: Access to undefined parameter `manifest_path` -- Initialise it to a default value eg. `params.manifest_path = some_value`
Missing `fromPath` parameter"

kbab: I am guessing  nextflow cannot find the configuration file. In your case is nextflow.config in the same directory as you are running the command from? I will amend readme to include `-c path_to_nextflow_config` to override this assumption.

AybukeKY (approved):
LGTM. Would be good to mention that one needs to install the bia-export package via "poetry install" first and then run the workflow.

PR #273, opened 2025-01-03
added accession number to sort parameters for export

sherwoodf:
Now correctly processes (and thefore will output) S-BIAD1480 before S-BIAD1407, even though they are released on the same data (was processing 1407 before 1480 resulting in the wonky order: https://deploy-preview-42--bia-beta-website.netlify.app/bioimage-archive/datasets/ )

Added info level log to help track how far through processing we are.

For some reason the file paths weren't working for the cache? They were last year... in any case i have updated them and they now work.

### DIFF
--- a/bia-export/bia_export/website_export/export_all.py
+++ b/bia-export/bia_export/website_export/export_all.py
@@ -5,6 +5,7 @@ from bia_export.bia_client import api_client
 from bia_integrator_api.models import Study
 from .generic_object_retrieval import read_api_json_file
 import logging
+import re
 
 logger = logging.getLogger("__main__." + __name__)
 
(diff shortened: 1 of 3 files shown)


bia-export/bia_export/website_export/export_all.py
```diff
@@ -39,15 +40,24 @@ def fetch_studies_from_api(
 
 
 def get_study_ids(root_directory: Optional[Path] = None):
+
+    def get_accno(acc_id):
+        match = re.search(r"\d+$", acc_id)
```
kbab, line 45: Is using the number alone to make things insensitive to the original collection? E.g. If we have a BSST study and a BIAD study? Or is it to prevent alphabetic sorting? (or both)?

kbab (approved):
LGTM - I only had one comment, happy for it to be merged as is.

bia-export/bia_export/website_export/export_all.py
sherwoodf: A bit of both, but mostly to prevent alphabetic e.g. S-BIAD1223 coming before S-BIAD992

PR #276, opened 2025-01-06
Better api logs

liviuba:
https://app.clickup.com/t/86979zne5

Everything for all requests in a single log format, might need to reconsider if logs get too big

LE: extra fields added should help tell if a slow request should be slow because it's just big, or if there is an issue somewhere


kbab (approved):
LGTM

PR #280, opened 2025-01-09
Add search/filter by uuid

liviuba:
https://app.clickup.com/t/8697a741r

search seemed like a better place since listing just by type independent of the actual model CRUD operations and model structure

This also deletes `getStudies` because it would be redundant now, with `searchStudy`


matthewh-ebi (approved):
This looks great, clear how it works and the tests are very useful in helping me understand what queries would look like. Excited to try it out :-)  

Am I correct in assuming that if we passed in the same UUID multiple times (e.g. we had something like filter_uuid [uuid1, uuid2, uuid3, uuid3, uuid3]), it ends up treated like a set (so the response would be [dataset1, dataset2, dataset3], not [dataset1, dataset2, dataset3, dataset3, dataset3])?

liviuba:
Yes, added it now as an extra testcase, thanks!

PR #282, opened 2025-01-13
get docker api & db running for export tests

sherwoodf:
https://app.clickup.com/t/8697a6f4e

PR adds:

- Added a make file and a new service in docker-compose for the project, to wire up the dependecy requirement of having a api running somewhere to run all the tests.
- Update github actions yaml to run export tests using the makefile
- Export client uses settings defined off env values, which makes switching the api client base url easy (plus it only affects things for the duration of testing)
- Export uses the search by accession ID to get study uuids, so that the (much easier to type/keep track of) accession ids can be passed in even when using the api
- Added an export all command, and updated the existing commands with subcommands to group them & make more sense
- Updated export readme
- Updated default export file naming, to be consistent with this (yet to be merged) pr: https://github.com/BioImage-Archive/BIA-astro/pull/43
- Update all the test files so that their contents is valid for sending to an up-to-date & empty API 

- Also updated the name of the readme for the client, as the github ci test was complaining about it - this has helped the Docker Compose Test API  action, but a new error is produced

Changes this PR doesn't cover:

- I wanted to create a generic function for 'throw these objects into the test api' - it's a little more complicated due to dependency ordering: https://app.clickup.com/t/8697fg88f
- Need to add more tests (will do so as part of 2nd PR)


bia-export/test/test_api_export.py
```diff
@@ -0,0 +1,75 @@
+from typer.testing import CliRunner
+from pathlib import Path
+from bia_export.cli import app
+import json
+
+runner = CliRunner()
+
+
+def test_cli_export_website_studies(tmp_path: Path, data_in_api):
```
liviuba, line 9: is `data_in_api` actually used somehwere?

bia-test-data/pyproject.toml
```diff
@@ -10,7 +10,9 @@ packages = [{include = "bia_test_data"}]
 [tool.poetry.dependencies]
 python = "^3.10,<3.12"
 bia-shared-datamodels = { path = "../bia-shared-datamodels", develop = true }
+bia-integrator-api = { path = "../clients/python", develop = true }
 pydantic-settings = "^2.3.4"
+python-on-whales = "^0.74.0"
```
liviuba, line 15: we don't need this dependency anymore

clients/README.md
```diff
@@ -0,0 +1,5 @@
+## Generating clients
```
liviuba, line 1: not sure why, but git says this (and the example readme) are new files

oh it's because it was renamed from Readme to README, I think we can just leave them out of the PR

liviuba (changes requested):
Looks good!

Tried running it and I get an error about comparison items being different on local. 
Happy to approve anyway if the actual testing part is meant to come in a different PR, but the maintainability-related comments would be good to change

bia-export/test/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+@pytest.fixture(scope="session")
+def data_in_api():
+    setttings = Settings()
+
+    input_file_dir = Path(__file__).parent / "input_data"
+
+    # Note order is *very* important due to uuid references needing to exist.
+    file_path_list = [
```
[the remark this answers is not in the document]
sherwoodf: Yep, there is definitely a better way. I've listed the order in the ticket in:  https://app.clickup.com/t/8697fg88f which should be doable.

clients/README.md
sherwoodf: I did this because the git CI was complaining about a lack of README as part of the cli test. I was hoping it would just fix the test but a new error message came up.

.github/workflows/core.yaml
```diff
@@ -39,3 +38,32 @@ jobs:
         run: poetry build
       - name: Run pytest
         run: poetry run pytest
+  ci-make:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11"]
+        poetry-version: ["1.4.2"]
+        os: [ubuntu-20.04]
```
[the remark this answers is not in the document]
sherwoodf: The mac and windows will fail due to docker daemons not running - it's not clear to me whether it's worth testing on all 3 (e.g. since mac ends up running on linux anyway) or what that would involve to get working....

bia-export/test/conftest.py
```diff
@@ -0,0 +1,132 @@
...
+import pytest
+from bia_test_data.data_to_api import add_objects_to_api
+from bia_export.settings import Settings
+from pathlib import Path
+import json
+import os
+
+
+def pytest_configure(config: pytest.Config):
+    if os.path.exists("/.dockerenv") or os.getenv("DOCKERIZED"):
```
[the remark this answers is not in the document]
sherwoodf: So this current behaviour is:

1. Env value is set (by file etc)
2. (if running tests, pytests overwrites env value)
3. Client uses env value

This was neat & simple when running tests only had 1 setup system, because there was only 1 override value. But now the override is different in the two cases i suspect it won't work. I could look into forwarding http://localhost:8080 to http://api:8080 in the docker version of the tests perhaps?

clients/README.md
sherwoodf: I know it's unrelated, and should have been it's own PR, but here's the logs from the previous runs:

https://github.com/BioImage-Archive/bia-integrator/actions/runs/12771730633/job/35599596262

and the latest ones (still error, but at a different points):

https://github.com/BioImage-Archive/bia-integrator/actions/runs/12793223210/job/35665541622?pr=282

liviuba:
> Looks good!
> 
> Tried running it and I get an error about comparison items being different on local. Happy to approve anyway if the actual testing part is meant to come in a different PR, but the maintainability-related comments would be good to change

Just tried now with no local db (everything deleted for a different ticket) and tests passed (so not worth investigating, just reusing-db-volume flakiness)

bia-export/test/conftest.py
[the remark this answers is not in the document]
liviuba: I think the `DOCKERIZED` flag is going to be the same kind of cheap-now-expensive-later thing as large blocks of code commented out, would something like a Settings class with a .env (somewhere, even in the .env_compose just with a previx for the variable name) work?

.github/workflows/core.yaml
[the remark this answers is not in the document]
liviuba: I'd rather keep them and fix the errors (or just explicitly ignore?). Just so at some point in the future we don't realise we've slightly diverged from being os-independent right when we need to be

clients/README.md
liviuba: Yes, I'd rather leave those just to figure out why are we getting the error now (probably to do with the poetry 2 release, but I'd rather see a changelog line or something)

liviuba:
To summarise, for suggested changes:
- [] data_in_api unused (but used a set-up)
- [] re-add dropped OS support (as in, tests that check os support)
- [] a settings-like class instead of `DOCKERIZED`

liviuba (approved):
Happy to merge, ty!

PR #283, opened 2025-01-13
Add stats

liviuba:
https://app.clickup.com/t/8697aa118

Mainly for the counts and total size - needs checking in prod (after merge) of whether we have issues with the result set after the first match

### DIFF
--- a/api/api/models/repository.py
+++ b/api/api/models/repository.py
@@ -7,7 +7,7 @@ from motor.motor_asyncio import (
 from enum import Enum
 import bia_shared_datamodels.bia_data_model as shared_data_models
 import pymongo
-from typing import Type, List, Any
+from typing import Type, List, Any, Dict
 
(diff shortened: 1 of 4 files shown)


api/api/models/repository.py
```diff
@@ -348,6 +348,12 @@ async def _get_docs_raw(self, pagination: Pagination, **kwargs) -> List[dict]:
 
         return docs
 
+    async def aggregate(self, pipeline):
```
sherwoodf, line 351: very much a nit pick: most of the other methods have return value type hints

sherwoodf (approved):
LGTM! I was going to comment on the variable called 'pipeline' because it didn't make sense to me initially, but i see it's what the AsyncIOMotorCollection calls it.

Can't run tests locally because mongo has stopped starting but looks like it passes.

PR #284, opened 2025-01-20
Fix no indices for local api

liviuba:
https://app.clickup.com/t/8697gahu4

To test:
(deletes everything on local)
```bash
docker stop $(docker ps -a -q)
docker rm $(docker ps -a -q)
docker volume rm $(docker volume ls -q)

docker compose --env-file ./.env_compose up --build
# check local mongo, bia-integrator db/collection should have ~19 indices
```


sherwoodf (approved):
Approved - not sure about the steps to test (i.e.  before the changes, there were still 19 indexes it seemed?) but i tested it with the code that orignially had the issue (the new docker tests) and the API correctly returned an error:

'TTP response body: {"detail":"Document a2fdbd58-ee11-4cd9-bc6a-f3d3da7fff71 already exists"

with the changes in this PR

PR #287, opened 2025-01-22
updated text expectations after pydantic version bump

sherwoodf:
Looking at the change, this shouldn't have any real effect - the objects now get turned to 'strings that are valid urls' when dumped & gain a trailing / if they didn't have one and don't have a fragment identifer.


liviuba (approved):
LGTM!

PR #288, opened 2025-01-23
fix typo and add comments

sherwoodf:
Issue: it looks like we kind of have overlapping use cases - technically bia-ingest is the thing that's always touching the data provided by submitters, but maybe we want to keep the distinction?

Ticket: https://app.clickup.com/t/8697d3tyd

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
@@ -18,10 +18,13 @@ class ConfiguredBaseModel(BaseModel):
 
 
 class AttributeProvenance(str, Enum):
-    submittor = "submittor"
+    # Additional information the user provided, though some submission system, that does not directly map to a field in our data models
+    submitter = "submitter"
 
+    # Information that was added by logic in the bia_ingest package, usually for use in downsteam submission processing code
     bia_ingest = "bia_ingest"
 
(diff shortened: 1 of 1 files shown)


bia-shared-datamodels/src/bia_shared_datamodels/mock_objects.py
```diff
@@ -492,7 +492,7 @@ def get_taxon_dict(completeness=Completeness.COMPLETE) -> dict:
 
 def get_attribute_dict(completeness=Completeness.COMPLETE) -> dict:
     attribute = {
-        "provenance": semantic_models.AttributeProvenance.submittor,
+        "provenance": semantic_models.AttributeProvenance.bia_ingest,
```
kbab, line 495: According to the comment above the "submitter" Enum definition it feels like this should be `semantic_models.AttributeProvenance.submitter` . I am happy to modify this in the ingest code for FileReferences, and you can just change the spelling of submittor -> submitter here. What do you think?

sherwoodf: This is kind of an generic/example attribute that gets used everywhere in the tests (i.e. every object with attributes), so i'm not sure it matters much - even though it says 'file_list_columns' in the name. I don't really mind what's here for the test as long as it passes. But since we don't use the submitter anywhere in our code, i question it's use in general.

kbab (approved):

PR #289, opened 2025-01-24
Fix examples install

liviuba:
https://app.clickup.com/t/8697gnfa3

bia-integrator-api built with poetry 2.0.1 in CI, 1.8.0 on my local (even on fresh builds) . removing all containers/images and rebuilding didn't fix it `docker system prune --all` did

Best guess is that poetry install layer kept being cached and used from the cache, but poetry was updated in the meanwhile so I was getting the stale version on my local only?

Similar fix to what @sherwoodf 's


sherwoodf (approved):
Yay it passes!

PR #290, opened 2025-01-24
added cli command that outputs uningested accession ids

sherwoodf:
ticket: https://app.clickup.com/t/8697na58z

Note this currently just uses the 'ingested_submissions' file as reference for what has been ingested. This should really use the API but i would want to update some code in the config.py & create clients slightly differently in order to make testing against an api easier. This seems like a separate ticket / would make the PR really big like the one for export, so splitting it up here.


Chr1st0p43rR (approved):
All good after duplicate test function name fix.

PR #291, opened 2025-01-24
Fix shared-model-tests back to original states

sherwoodf:
pydantic changed url serialization back in https://docs.pydantic.dev/latest/changelog/#v2106-2025-01-23, so returning tests to original state


liviuba (approved):
LGTM!

PR #292, opened 2025-01-27
Minor improvements to export code

sherwoodf:
updated README, default image file name to be the same as that in astro, and minor function udpate


kbab (approved):
LGTM

PR #293, opened 2025-01-28
added dependency based ordering to add_objects_to_api for testing

sherwoodf:
Ticket: https://app.clickup.com/t/8697fg88f

Functionality allows us to just throw a bunch of files at the api (e.g. during test setup) and they will get created in the correct order.

Errors will be thrown if something is dodgy with the chain of dependencies

### DIFF
--- a/bia-export/test/conftest.py
+++ b/bia-export/test/conftest.py
@@ -18,7 +18,7 @@ def pytest_configure(config: pytest.Config):
 def data_in_api():
     setttings = Settings()
 
-    input_file_dir = Path(__file__).parent / "input_data"
+    input_file_dir = Path(__file__).parent / "input_data" / "**" / "*.json"
     file_path_list = glob(str(input_file_dir), recursive=True)
 
     object_list = []
--- a/bia-export/test/conftest.py
+++ b/bia-export/test/conftest.py
@@ -26,10 +26,6 @@ def data_in_api():
         if os.path.isfile(path):
             with open(path, "r") as object_file:
                 json_dict = json.load(object_file)
-                try:
-                    json_dict["model"]["type_name"]
-                except KeyError:
-                    continue
                 object_list.append(json_dict)
 
     add_objects_to_api(setttings.api_base_url, object_list)
--- a/bia-export/test/conftest.py
+++ b/bia-export/test/conftest.py
@@ -6,6 +6,7 @@ from bia_export.settings import Settings
 from pathlib import Path
 import json
 import os
+from glob import glob
 
 
 def pytest_configure(config: pytest.Config):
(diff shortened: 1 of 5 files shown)


bia-test-data/bia_test_data/data_to_api.py
```diff
@@ -29,7 +29,7 @@ def get_object_creation_client(
             username=user_dict["email"], password=user_dict["password_plain"]
         )
     except exceptions.UnauthorizedException:
-        private_api.register_user(user_dict)
+        private_api.register_user(api_models.BodyRegisterUser(**user_dict))
```
sherwoodf, line 32: unrelated - just making VSCode happy.

bia-export/test/conftest.py
```diff
@@ -17,114 +18,18 @@ def pytest_configure(config: pytest.Config):
...
-                json_dict["model"]["type_name"]
-            except KeyError:
-                continue
-            object_list.append(json_dict)
+    for path in file_path_list:
+        if os.path.isfile(path):
+            with open(path, "r") as object_file:
+                json_dict = json.load(object_file)
+                try:
+                    json_dict["model"]["type_name"]
```
liviuba, line 30: This looks to me like a way to handle jsons that are in the same directory tree as things to be pushed into the api but don't look like they're typed with shared models - so I don't understand why would they be in the directory tree in the first place (and need to be skipped)?

bia-test-data/bia_test_data/data_to_api.py
```diff
@@ -41,12 +41,176 @@ def get_object_creation_client(
     return private_api
 
 
-def add_objects_to_api(api_base_url, object_list=list[dict]):
+def calculate_dependency_chain_length(
```
liviuba, line 44: I don't think doing it this way is bad on principle, but I'm thinking of when we'll change the models eventually. Suppose we add a new model type that would have dependencies, is there any way that code that uses this would silently ignore the new model (instead of e.g. just failing so we'd know we need to change this too)?

liviuba (approved):
LGTM! First comment is something I'd change or add a comment for

bia-export/test/conftest.py
sherwoodf: Ah yep, i think we can get rid of this check here.

bia-test-data/bia_test_data/data_to_api.py
sherwoodf: So we could make this document based rather than model based. This would mean that all files we sent went through the same, more complicated, system. Perhaps that would be better (would mean there is only ever 1 case to debug, but at least the debugging is always on the same system). It would probably follow a similar 'dependency chain length calculation'

If we added new models and added tests (since this is part of the test package) then the objects wouldn't make it to the API, so the tests would fail if they needed objects in the API. Which feels safe enough to me.

sherwoodf: Let me know if you think it would be worth changing it to this!

PR #294, opened 2025-01-29
created attribute subclasses for used across BIA packages

sherwoodf:
Possible approach to attribute subclassing to create a single location / contract for other BIA packages to use.

Tested against local api using S-BIAD1285 - but would be good to have proper API testing for ingest to make sure it works - but not in scope of this PR. 

```
-------- Starting ingest of S-BIAD1285 --------
INFO     Fetching submission from https://www.ebi.ac.uk/biostudies/api/v1/studies/S-BIAD1285                                                                    api.py:229
INFO     Created Affiliation. Count: 1                                                                                                                  cli_logging.py:180
INFO     Created Affiliation. Count: 1                                                                                                                  cli_logging.py:180
INFO     Created Contributor. Count: 5                                                                                                                  cli_logging.py:180
INFO     Created FundingBody. Count: 1                                                                                                                  cli_logging.py:180
INFO     Created Grant. Count: 1                                                                                                                        cli_logging.py:180
INFO     Created Study. Count: 1                                                                                                                        cli_logging.py:180
INFO     Created ImageAcquisitionProtocol. Count: 1                                                                                                     cli_logging.py:180
INFO     Created AnnotationMethod. Count: 1                                                                                                             cli_logging.py:180
INFO     Created SpecimenImagingPreparationProtocol. Count: 1                                                                                           cli_logging.py:180
INFO     Created Protocol. Count: 0                                                                                                                     cli_logging.py:180
INFO     Created Taxon. Count: 1                                                                                                                        cli_logging.py:180
INFO     Created BioSample. Count: 1                                                                                                                    cli_logging.py:180
INFO     Created ImageAnalysisMethod. Count: 0                                                                                                          cli_logging.py:180
INFO     Created ImageCorrelationMethod. Count: 0                                                                                                       cli_logging.py:180
INFO     Created Dataset. Count: 2                                                                                                                      cli_logging.py:180
INFO     Skipping file reference creation.                                                                                                     process_submission_v4.py:75
-------- Completed ingest of S-BIAD1285 --------
```

Would also be good to improve the errror/creation tracking  counts so we don't have to keep adding fields for objects - but that is definitely not in scope.

### DIFF
--- a/bia-export/bia_export/website_export/studies/retrieve.py
+++ b/bia-export/bia_export/website_export/studies/retrieve.py
@@ -9,7 +9,7 @@ from bia_export.website_export.generic_object_retrieval import (
 )
 from pathlib import Path
 from pydantic import ValidationError
-from pydantic.alias_generators import to_snake, to_pascal
+from pydantic.alias_generators import to_snake
 from .models import StudyCLIContext, CacheUse
 from bia_shared_datamodels import bia_data_model, attribute_models
 from bia_integrator_api import models as api_models
--- a/bia-export/bia_export/website_export/studies/retrieve.py
+++ b/bia-export/bia_export/website_export/studies/retrieve.py
@@ -8,9 +8,10 @@ from bia_export.website_export.generic_object_retrieval import (
     retrieve_object,
 )
 from pathlib import Path
-from pydantic.alias_generators import to_snake
+from pydantic import ValidationError
+from pydantic.alias_generators import to_snake, to_pascal
 from .models import StudyCLIContext, CacheUse
-from bia_shared_datamodels import semantic_models, bia_data_model
+from bia_shared_datamodels import bia_data_model, attribute_models
 from bia_integrator_api import models as api_models
 import json
 from typing import List, Type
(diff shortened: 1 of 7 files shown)


bia-export/bia_export/website_export/studies/retrieve.py
```diff
@@ -8,9 +8,10 @@
     retrieve_object,
 )
 from pathlib import Path
-from pydantic.alias_generators import to_snake
+from pydantic import ValidationError
+from pydantic.alias_generators import to_snake, to_pascal
```
sherwoodf, line 12: to_pascal is not needed.

bia-ingest/bia_ingest/biostudies/generic_conversion_utils.py
```diff
@@ -16,6 +16,7 @@
 from ..config import settings, api_client
 from ..cli_logging import IngestionResult, log_failed_model_creation
 import bia_integrator_api.models as api_models
+from bia_shared_datamodels import attribute_models
```
sherwoodf, line 19: This is actually not necessary

bia-ingest/bia_ingest/biostudies/v4/dataset.py
```diff
@@ -254,7 +258,10 @@ def store_annotation_method_in_attribute(
         raise RuntimeError(
             "Dataset cannot find Annotation Method that should have been created"
         )
-    return attribute_dicts
+    return [
```
kbab, line 261: This seems to return objects rather than dicts. Downstream it does not seem to matter when the dataset is persisted though.

kbab (commented):
The attribute models enforce structure at the cost of slightly more complex code, but I think the effect is positive.

matthewh-ebi (approved):
This would be very useful for the curation code if we decide to enforce a structure that includes more detailed provenance information through Attribute structure. Would be interesting to test that approach and see how we get on.

kbab:
A note before merging, for export are we going to cater for affected attributes in studies ingested before this or modify those attributes in the API to use the new models?

sherwoodf:
> A note before merging, for export are we going to cater for affected attributes in studies ingested before this or modify those attributes in the API to use the new models?

I believe this is meerly documenting existing attributes, and not editing them, so there should be no effect.

PR #295, opened 2025-01-29
Update instances using deprecated api_client method 'get_studies'

kbab:
PR #280  deprecated the use of the `get_studies` method of the API. This PR updates the two places it was being used in the code base. [Clickup ticket ](https://app.clickup.com/t/8697qnjhj)


sherwoodf (approved):
lgtm!

PR #298, opened 2025-01-31
updated input files with newly ingested studies

sherwoodf:
After running ingest for: https://app.clickup.com/t/8697qxxjf

Note: S-BIAD1515, S-BIAD1524 produced errors - seem to be related to pagetab processing.


Chr1st0p43rR (approved):
All looks good! 

Just one thing to clarify — if I run your `biaingest find new-biostudies-studies`, I get 5 more entries in the "uningested_studies_of_interest_<today's date>" than in the already-there uningested_studies_of_interest" — they're just new ones, is that right?

PR #299, opened 2025-01-31
update export to use new dataset stats fields

sherwoodf:
ticket: https://app.clickup.com/t/8697rdjuz


bia-export/README.md
Chr1st0p43rR, line 1: I wonder if it is worth correcting this part of the readme before merging — this:

To test that installation has worked correctly, you can run:
    poetry run bia-export website all S-BIADTEST -r test/input_data
    
just needs a space so that it is formatted correctly.

Chr1st0p43rR (approved):
All looks good — I noticed, following the readme, that the start of it (which isn't actually a change in this PR) could be corrected.

bia-export/README.md
sherwoodf: Fixed

PR #302, opened 2025-02-07
Assign images with proposals

matthewh-ebi:


bia-assign-image/bia_assign_image/cli.py
```diff
@@ -239,6 +242,52 @@ def create(
...
+            create(
+                accession_id=p['accession_id'],
+                image_uuid_list=[image_uuid],
+                persistence_mode=persistence_mode
+            )
+
+
+@app.command(help="Propose file references to convert for an accession")
+def propose_files(
+    accession_id: Annotated[str, typer.Argument(help="The accession ID to process")],
```
kbab, line 276: Is this intentionally restricted to accept just one  accession ID?

kbab (commented):
Happy to merge as is. However, Readme needs updating. Happy to copy relevant section from bia-converter-light and commit then approve.

bia-assign-image/bia_assign_image/cli.py
matthewh-ebi: Good point, I think I'm used to other tools which work with single IDs but no reason it could allow multiple.

kbab (approved):
LGTM

PR #304, opened 2025-02-10
test new way of wiring up tests

sherwoodf:
Ticket: https://app.clickup.com/t/8697kxtgx

Scope creep: added a couple of tests that would have caught the recent changes to the api

### DIFF
--- a/.github/workflows/core.yaml
+++ b/.github/workflows/core.yaml
@@ -65,13 +65,11 @@ jobs:
         with:
           poetry-version: ${{ matrix.poetry-version }}
       - name: Run docker
-        run:  docker compose up -d --build   
+        run:  docker compose up -d --build --wait   
       - name: set up poetry
         run: |
           poetry env use python
           poetry install
-      - name: Check to see if poetry can build
-        run: poetry build
       - name: Run pytest
         run: poetry run pytest
       # Always cleanup - even for cancelled jobs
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: api.version client.generate bia-export.test
+.PHONY: api.version client.generate api-for-testing
```
liviuba, line 1: why the `api-for-testing` phony here?

.github/workflows/core.yaml
```diff
@@ -38,33 +38,45 @@ jobs:
...
-      - name: run makefile
-        run: make ${{matrix.project}}.test
+          poetry-version: ${{ matrix.poetry-version }}
+      - name: Run docker
+        run:  docker compose up -d --build   
+      - name: set up poetry
+        run: |
+          poetry env use python
+          poetry install
+      - name: Check to see if poetry can build
```
liviuba, line 73: why the build here we're not packaging export?

bia-test-data/bia_test_data/data_to_api.py
```diff
@@ -41,6 +43,26 @@ def get_object_creation_client(
     return private_api
 
 
+def get_client_with_retries(
```
liviuba, line 46: if we're taking this approach, I'd rather the api start command would wait until the api is up so clients that use it as a dependency don't have to wait themselves or care about synchronisation

liviuba (commented):
It's good to have tests run on the host vs in containers all the time!

Not entirely sure about how service dependencies are set up though

liviuba (approved):
LGTM!

PR #308, opened 2025-02-14
Export update command

sherwoodf:
Ticket: https://app.clickup.com/t/8697wq389

Also removed the unnecessary cache for dataset stats now that the api has the great endpoints.

tested locally with:
```
poetry run bia-export website study S-BIAD830 -o bia-updated.json -u bia-study-metadata.json
```
It's super fast! Note you have to point at your own copy of study metadata (e.g. the one stored in astro)

Currently doesn't overwrite data unless you have -o == -u file paths (which can also happen if -u points to the default file)

I'm not super happy with the test i wrote - i think it could be done in a neater way, but it's good enough for now. 
Note, not actually testing update for the image & image dataset functions - their logic is a lot lot simpler than the study one though, because they don't care about order.


AybukeKY (approved):
LGTM though I couldn't run pytest due to my api issues.

PR #311, opened 2025-02-25
added test for bia agent 

sherwoodf:
and updated some ingest test wiring and how settings get used

Changing settings = Settings() to a get_settings method allows lazy-loading, making it less of a headache to set env variables and then run tests.

Made sure the test for find studies uses local api (was using the actual wwwdev. one)

https://app.clickup.com/t/8697wpbjk


AybukeKY (approved):
LGTM.

PR #312, opened 2025-03-04
Fix bug in propose resulting in same file being proposed more than once

kbab:
A very small PR to fix bug in propose code resulting in same file being proposed more than once. [Clickup ticket](https://app.clickup.com/t/86986kxn6)


sherwoodf (approved):
LGTM!

PR #313, opened 2025-03-07
Make propose read and write to yaml as well as tsv

kbab:
Allow yaml format to be used in addition to tsv format for proposing file references to convert to assign to images - [clickup ticket](https://app.clickup.com/t/86985ruc6)


sherwoodf (approved):
lgtm!

PR #315, opened 2025-03-18
Update elastic indexed fields

liviuba:
https://app.clickup.com/t/8698a6xhc

Seems ok for our usecase but might need to revisit and be clever with index generation.
TL;DR: we always want `"dynamic": false,`

### DIFF
--- a/api/api/tests/test_data/Readme.md
+++ b/api/api/tests/test_data/Readme.md
@@ -18,7 +18,7 @@ curl -X PUT "http://localhost:9200/test-index" \
 			},
 			"dataset": {
 				"type": "flattened"
-			}
+			},
             "title": {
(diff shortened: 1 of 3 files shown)


api/api/tests/test_data/Readme.md
```diff
@@ -16,9 +16,15 @@ curl -X PUT "http://localhost:9200/test-index" \
 			"author": {
 				"type": "flattened"
 			},
+			"dataset": {
+				"type": "flattened"
+			}
             "title": {
```
sherwoodf, line 22: I think this line is missing a comma to be valid json

sherwoodf (approved):
Approving but not merging due to investigation on json going into the query.

PR #316, opened 2025-03-18
Bia converter with docker

kbab:
Make image conversion prefer to use official ome bioformats2raw docker container [https://app.clickup.com/t/8698ad224] (https://app.clickup.com/t/8698ad224) .


bia-converter/tests/test_running_bioformats2raw.py
```diff
@@ -0,0 +1,42 @@
...
+@pytest.fixture(scope="function")
+def output_dirpath(tmp_path) -> Path:
+    """Return path to which converted zarr should be saved"""
+
+    return tmp_path / "converted.zarr"
+
+
+@pytest.mark.parametrize(
+    "bioformats2raw_function",
+    (run_bioformats2raw_with_docker, run_bioformats2raw_with_singularity),
```
liviuba, line 32: it's odd to me to have both options here, because one would usually fail if the other one passes - like [here](https://github.com/BioImage-Archive/bia-integrator/actions/runs/13934847886/job/39000407311?pr=316)

Is there a way to test *some* of the functionality, but still have the test pass in CI?

kbab: Singularity test was to ensure singularity version works - important for codon. Put a `skipif ` to skip test if singularity not available on machine.

Chr1st0p43rR:
For posterity following our conversation — test fails at retrieving docker image with Mac ARM64.

kbab:
> For posterity following our conversation — test fails at retrieving docker image with Mac ARM64.

Thanks. Added conditional check to use java cli for this case.

liviuba (approved):
LGTM!

Chr1st0p43rR (approved):
Looks good — checked new tests, worked.

PR #317, opened 2025-03-24
Enable auth for elastic, add alternative viewer

liviuba:
https://app.clickup.com/t/8698a6z8h

Adds auth and an alternative viewer because keeping kibana meant either adding multi-step setup with an auth token or setting up separate role configs (for dev) - all overkill, might need to reconsider


sherwoodf (approved):
lgtm

PR #321, opened 2025-04-04
Add script to refresh elastic index

liviuba:
https://app.clickup.com/t/8698a755k

To test: start the api and run the script, all variable defaults should match the ones in docker compose.
Uses api already configured for export, might revisit after using it for a while to cleanup environment


sherwoodf (approved):
Works for me! looks good. Might be worth pointing to it in some readme, but not sure which.

PR #322, opened 2025-04-09
Add Elastic config to api helm chart

liviuba:
https://app.clickup.com/t/8698u9x12


api/Readme.md
```diff
@@ -123,4 +123,4 @@ Install act (Ubuntu):
 Run:
 * Run everything: `gh extension exec act`
 * List workflows for an event: `gh extension exec act -l pull_request`
-* Run api tests `gh extension exec act push -j docker-compose-test --secret-file act_secret.env --var-file act_vars.env`
\ No newline at end of file
+* Run api tests `gh extension exec act push -j docker-compose-test --secret-file act_secret.env --var-file act_vars.env`
```
kbab, line 126: I can't run this locally. I get an error `Error: stat /home/kola/code/embl_projects/BioImageArchive/bia-integrator/api/.github/workflows: no such file or directory `. Is it ok to ignore this?

api/api/app.py
```diff
@@ -109,9 +109,7 @@ async def on_start():
 async def on_stop():
     event_loop = asyncio.get_event_loop()
     app.extra["extra"]["event_loop_specific"][event_loop]["db"].connection.close()
-    await app.extra["extra"]["event_loop_specific"][event_loop][
-        "elastic"
-    ].client.close()
+    await app.extra["extra"]["event_loop_specific"][event_loop]["elastic"].close()
```
kbab, line 112: There is a slight difference in the command (.client.close vs .close) is this important?

kbab (commented):
LGTM and happy to approve. I got errors running tests from `bia-integrator/api` but those from `bia-integrator/api/api` worked. Also left comment about running with github cli commands.

api/api/app.py
liviuba: yes, there is also a change in the elastic handler that closes only if the client is non-null (part of this PR that allows it to be initialised with a null client as a noop if not set)

kbab (approved):
LGTM

PR #323, opened 2025-04-11
Fix minor bug when updating version during persistence using API

kbab:
Clickup ticket: [https://app.clickup.com/t/8698pbpje](https://app.clickup.com/t/8698pbpje)

### DIFF
--- a/bia-assign-image/bia_assign_image/api_client.py
+++ b/bia-assign-image/bia_assign_image/api_client.py
@@ -99,9 +99,12 @@ def store_object_in_api_idempotent(api_client, model_object):
         return
 
     if equivalent_api_object == api_copy_of_object:
+        logger.info(
+            f"Not persisting current object as identical object {model_name} with UUID {model_object_uuid} already exists in API."
+        )
         return
     else:
-        model_object.version = api_copy_of_object + 1
+        equivalent_api_object.version = api_copy_of_object.version + 1
 
         logger.info(f"Updating {model_name} with UUID {model_object_uuid} in API")
-        post_func(model_object)
+        post_func(equivalent_api_object)
--- a/bia-assign-image/bia_assign_image/api_client.py
+++ b/bia-assign-image/bia_assign_image/api_client.py
@@ -101,7 +101,7 @@ def store_object_in_api_idempotent(api_client, model_object):
     if equivalent_api_object == api_copy_of_object:
         return
     else:
-        model_object.version += 1
+        model_object.version = api_copy_of_object + 1
 
         logger.info(f"Updating {model_name} with UUID {model_object_uuid} in API")
         post_func(model_object)
--- a/bia-assign-image/tests/test_store_object_in_api_idempotent.py
+++ b/bia-assign-image/tests/test_store_object_in_api_idempotent.py
@@ -5,7 +5,6 @@ import pytest
 
 from bia_shared_datamodels import bia_data_model, uuid_creation
 from bia_integrator_api import PrivateApi
-from bia_integrator_api.exceptions import NotFoundException
 
 from bia_assign_image.api_client import (
     get_local_bia_api_client,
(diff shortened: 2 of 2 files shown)


liviuba (commented):
Good to ensure the write would be successful, just some minor unclarities to address

bia-assign-image/bia_assign_image/api_client.py
```diff
@@ -101,7 +101,7 @@ def store_object_in_api_idempotent(api_client, model_object):
     if equivalent_api_object == api_copy_of_object:
         return
     else:
```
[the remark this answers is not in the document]
kbab: Added tests.

bia-assign-image/tests/test_store_object_in_api_idempotent.py
```diff
@@ -0,0 +1,113 @@
...
+    api_copy_of_new_study = api_client.get_study(new_study_uuid)
+    assert compare_bia_study_object_with_api_study_object(
+        new_study_object, api_copy_of_new_study
+    )
+
+
+def test_store_exact_same_object_returns_object_exists_message(
+    new_study_uuid, new_study_object, api_client, caplog
+):
+    # Ensure the object is not in the API
```
liviuba, line 65: I don't think we need to test the fixtures for correctness

bia-assign-image/tests/test_store_object_in_api_idempotent.py
```diff
@@ -0,0 +1,113 @@
...
+def new_study_uuid(new_study_object) -> str:
+    return str(new_study_object.uuid)
+
+
+@pytest.fixture
+def api_client() -> PrivateApi:
+    return get_local_bia_api_client()
+
+
+def compare_bia_study_object_with_api_study_object(bia_study, api_study):
```
liviuba, line 40: Is this meant to test something that's part of this PR?

bia-assign-image/tests/test_store_object_in_api_idempotent.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # First save
+    store_object_in_api_idempotent(api_client, new_study_object)
+    api_copy_of_new_study = api_client.get_study(new_study_uuid)
+    assert compare_bia_study_object_with_api_study_object(
+        new_study_object, api_copy_of_new_study
+    )
+
+    # Second save
+    with caplog.at_level(logging.INFO):
```
liviuba, line 77: why are we doing this?

liviuba (commented):
The initial comment was meant to address `equivalent_api_object.version = api_copy_of_object.version + 1` missing the second `.version`, which is now fixed thanks to the tests.
Tests usefully introduce some patterns that are new and very enthusiastic, but have some verbose parts

bia-assign-image/tests/test_store_object_in_api_idempotent.py
kbab: Will remove these from tests.

bia-assign-image/tests/test_store_object_in_api_idempotent.py
kbab: This is to verify that the conditional branch of the code runs when attempting to persist exact same object to api. Will add comment about this.

bia-assign-image/tests/test_store_object_in_api_idempotent.py
```diff
@@ -0,0 +1,110 @@
...
+        new_study_object, api_copy_of_new_study
+    )
+
+    # Second save of exactly the same object
+    with caplog.at_level(logging.INFO):
+        store_object_in_api_idempotent(api_client, new_study_object)
+
+    # Check function logs info message that object is not being
+    # persisted because it is exactly the same as what is in API
+    expected_info_text = f"Not persisting current object as identical object Study with UUID {new_study_uuid} already exists in API."
```
liviuba, line 79: Is there a more robust way to check the function works as intended?

bia-assign-image/tests/test_store_object_in_api_idempotent.py
kbab: The assertion on line 84 checks that the store operation is idempotent. Capturing the log is to ensure that the intended logic is followed - it is not strictly necessary, as long as the idempotent assertion passes. Happy to remove the `with` block and simply test idempotency.

liviuba: In general, relying on logs isn't considered a robust way to test functionality

kbab: I have removed the check of the logs from the test. The assertion tests the functionality that the save is idempotent ...

liviuba (approved):
:+1:

PR #324, opened 2025-04-15
model updates

sherwoodf:
Ticket: https://app.clickup.com/t/8698nhx0n

Update model diagram: https://docs.google.com/drawings/d/1QWKamhDh4-3R1F2nJNR5M-Se-CXZxA6qWn58Qb6l8EE/edit


liviuba:
Some notes after a 1st read (numbered for ease to reply, not importance), but will keep looking at the change:

1. Why did we go from null being allowed vs not allowed for lists specifically? Especially in the context of interoperability, how would we model "Not available" and "Not provided" (/do we explicitly want not to?)
2. Inconsistent naming for `ProvenanceType` vs `Licence` with Type dropped in this PR 
3. For Licence, enum values needed to be human-readable because they were being used (e.g.) as the label for a dropdown item the user would select. How are we going to separate labels and values in the future?
4. How is ProvenanceType meant to interact with the `*_metadata` fields?
5. Just noticed the `title_id` now, I can see it's changed but how did we get to that field name in the first place? We need to make sure the initial purpose of the field is maintained for downstream code
6. (less important but still rename) biosample->bio_sample , this as a standalone change would be breaking with minor benefit, why change it here?
7. Do we have any usecases for efo_type_id being useful for search?

This looks like a PR that solves multiple pain points, some explicit some not. Organising the changes into different independent PRs would likely make them more cohesive and logical at the individual PR level

sherwoodf:
1. I believe we first encountered this as an issue where Shared Data Model package =/= Fast API client model - or at least the assumptions made in BIA ingest and EMPIAR ingest were different, resulting in changes being required in export when EMPIAR ingest started. We (as in you, me, matthew) all agreed that the option to have nullable as well as empty lists was confusing for client code. I'm not sure about "not avaliable". Not provided, vs explicitly provided as empty is maybe useful, and is i guess handled in non-list situations by the disctinction between: null, "", and "VALUE" for e.g. an optional string field.
2. Yeah i kinda agree
3. I would argue that they're a) still human readable enough to be read in the json and b) any situation involving a dropdown menu of options would also be able to handle a mapping between said values and something more human readable.
4. A good question, but from discussion just now with Matthew, suggestion would be that the two metadata fields would actually be better merged into a single field for the majority of our user base - which i think kind of makes the question no longer relevant i would guess?
5. The field name was title_id because in biostudies it was used both as a title but also an ID to reference the object in associations, thus the data (in biostudies) was attempting to capture both, and the mixin class indicated objects that the user touched in biostudies. However,  this a much to biostudies-centric approach that doesn't make an sense even when dealing with empiar ingest, and the data, once ingested into the API, is only used to provide the title of sections on the website. The original data is used in bia-ingest to provide the links between datasets and protocols + biosamples, but this isn't the case once the data is in the API.
6. Purely consistency with the other mock object function naming.
7. Searching for any studies where X kind of protocol occurred (say, studies with a HCS library protocol?) It would be similar to wanting to search for studies with an AnnotationMethod, but give us the potential option to have useful facets.

sherwoodf:
Closing in favour of smaller set of PRs

PR #325, opened 2025-04-16
uuid creation updates

sherwoodf:
Ticket: https://app.clickup.com/t/8698nhwmm

Doc: https://docs.google.com/document/d/1TxkOvf7z24OR90ljVVIhXDbK2r-a7S7OVNraYImzp0A/edit?tab=t.qieduw1hg7fs

Note, branch is set to be merged into Model update branch

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/uuid_creation.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/uuid_creation.py
@@ -1,7 +1,5 @@
 from uuid import UUID
 import hashlib
-from pathlib import Path
-from typing import Optional
 from .bia_data_model import (
     Dataset,
     FileReference,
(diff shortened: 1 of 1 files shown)


kbab (commented):
LGTM - happy to approve.  I was wondering if having a separate function for creating the uuid of each class was deliberate as the modification makes it possible to have the same signature for each function, so in theory we could have one function i.e. `create_xxx_uuid(study_id: str, class_name: str, unique_string: str)`

sherwoodf:
Ah yes, that's what i was thinking of doing, but then we wouldn't be setting class_name ourselves. It would probably fine - but i would just have this tiny concern about issues around passing in shared_data_model classes vs api client classes (which should have the same name, but the classes are not 'the same' in terms of pydantic / class equivalence, even if they have all the same fields).

PR #326, opened 2025-04-17
change licence enum to use full deed url

sherwoodf:
We currently store an enum string like: CC_BY_4.0. If we were to instead store the url to the spec (i.e. https://creativecommons.org/licenses/by/4.0/ ) this would:
-  Be more interoperable across repositories, since the URI would the the same anywhere, but any string version of it is defined only by us.
-  Require fewer changes/mapping when it comes to ro-crate data - RO crate data requires a URI of a licence (note we still do some mapping from biostudies anyway, because there are typos etc)
- Doesn't allow us to use the value for more human readable display, but there is some formatting that still needs doing anyway (removing underscores) & this way we can also provide links out to the licences.
- I'm keeping it as an enum, to restrict which licences were available: which is the main benefit of having an enum.

Changed the name to Licence - this is clearer about the intent of what it contains (the URIs are just the licences themselves).


liviuba (approved):
:+1:

PR #327, opened 2025-04-17
Make all Optional[List[]] fields only List[]

sherwoodf:
Ticket: https://app.clickup.com/t/8698nhx0n

This reason for this PR was due to one of the issues encountered during getting EMPIAR data into the API / switching from the shared data models to the API models. Something should probably be changed to avoid issues, this solution is probably 1 of 2-3 options.

As an example, comparing the attribute field and the file_uri in image representations:

Generated models: 
- https://github.com/BioImage-Archive/bia-integrator/blob/main/clients/python/bia_integrator_api/models/image_representation.py

With fields:
-     attribute: Optional[List[Attribute]] = None
-     file_uri: List[StrictStr] = Field(description="URI(s) of the file(s) which together make up this image representation.")

Shared models: 
- https://github.com/BioImage-Archive/bia-integrator/blob/main/bia-shared-datamodels/src/bia_shared_datamodels/mock_objects.py#L175
and 
- https://github.com/BioImage-Archive/bia-integrator/blob/main/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py#L313

With fields: 
- attribute: Optional[list[Attribute]] = Field(
        default_factory=list,
        description="""Freeform key-value pairs from user provided metadata (e.g. filelist data) and experimental fields.""",
    )
- file_uri: List[str] = Field(
        description="""URI(s) of the file(s) which together make up this image representation."""
    )

What this means is that the mock objects (and shared model tests, and anything using the shared models) expect/create a default to be created for these of [], but the API models create a default on null. Export ran into issues because BioStudies ingest used the shared models, and followed the expectations there, but EMPIAR ingest used client models - so the code had to be updated to also deal with all the null values.

A few extra points:
- I did not realise that the client creates models for sub-objects as well as the main objects! (E.g. Contributor has a class despite not being a root document)
- I did not realise that we have old models in there (e.g. experimentally_captured_image) we should delete these, but that's for a separate task.
- We should in general only use the client models, but I think that the expectations around the mock object should be better aligned with the client models

In order to resolve this potential confusion we could:

1. Make list fields non-optional. I believe this would update fastAPI to use the default_factory function. This would make our json objects easier to use: currently the website has to code defensively against both 'null' and '[]' to deal with situations where the is no data. I believe this would generally be a non-breaking change because we're just removing one possible option.
2. Make the default in the model for optional list fields None, rather than using default_factory. This would make the shared models more closely aligned with the client models. This also retains the expressiveness of 'null' vs '[]' as two different values - arguably this is similar to having checks for 'null' vs '""' for Optional strings. Our code has already been updated to deal with this case.
3. Do something to wrangle the fast api client generators to follow the shared models? (this seems unnecessarily complex, but including for completeness)

As someone who was writing the javascript for the website, I was leaning towards option 1 to make my code cleaner & general have fewer cases to worry about. I also kind of think that the promise of capturing 'user didn't provide a value for the field' and 'user intentionally provided a value that was null' rarely works out in practice & we still have to deal with cases where people fill in text fields with "N/A".

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/bia_data_model.py
@@ -255,22 +255,22 @@ class CreationProcess(semantic_models.CreationProcess, DocumentMixin):
         description="The biological specimen that is the subject of the image.",
     )
     image_acquisition_protocol_uuid: Annotated[
-        Optional[List[UUID]], ObjectReference(ImageAcquisitionProtocol)
+        List[UUID], ObjectReference(ImageAcquisitionProtocol)
     ] = Field(
         default_factory=lambda: [],
         description="The imaging protocol, describing the technique that was used to create the image.",
     )
 
-    input_image_uuid: Annotated[Optional[List[UUID]], ObjectReference(Image)] = Field(
+    input_image_uuid: Annotated[List[UUID], ObjectReference(Image)] = Field(
         default_factory=lambda: [],
         description="The images used as input data for the creation of a new image.",
     )
-    protocol_uuid: Annotated[Optional[List[UUID]], ObjectReference(Protocol)] = Field(
+    protocol_uuid: Annotated[List[UUID], ObjectReference(Protocol)] = Field(
         default_factory=lambda: [],
         description="A protocol which was followed that resulted in the creation of this new image from existing image data.",
     )
     annotation_method_uuid: Annotated[
-        Optional[List[UUID]], ObjectReference(AnnotationMethod)
+        List[UUID], ObjectReference(AnnotationMethod)
     ] = Field(
         default_factory=lambda: [],
         description="The annotation method describing the process followed to create a new image from exsiting image data.",
--- a/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/semantic_models.py
@@ -44,7 +44,7 @@ class AttributeMixin(BaseModel):
     Mixin for just the attribute field
     """
 
-    attribute: Optional[list[Attribute]] = Field(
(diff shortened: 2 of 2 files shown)


liviuba (approved):
:+1:

matthewh-ebi:
Yep, it's a bit of extra client work, but I think default empty lists is cleaner. We can then avoid this messy thing in the EMPIAR ingest code: https://github.com/BioImage-Archive/ingest-from-empiar/blob/49a45d21c34281cc7f19184f2aa2446cfeb8a54a/empiar_ingest/utils.py#L232

matthewh-ebi:
Yep, it's a bit of extra client work, but I think default empty lists is cleaner. We can then avoid this messy thing in the EMPIAR ingest code: https://github.com/BioImage-Archive/ingest-from-empiar/blob/49a45d21c34281cc7f19184f2aa2446cfeb8a54a/empiar_ingest/utils.py#L232

matthewh-ebi:
Yep, it's a bit of extra client work, but I think default empty lists is cleaner. The EMPIAR ingest code now has workarounds to force conversion of fields initialised objects that are optional lists to be empty lists instead, and we could then remove those.

matthewh-ebi:
Yep, it's a bit of extra client work, but I think default empty lists is cleaner. We can then avoid this messy thing in the EMPIAR ingest code: https://github.com/BioImage-Archive/ingest-from-empiar/blob/49a45d21c34281cc7f19184f2aa2446cfeb8a54a/empiar_ingest/utils.py#L232

matthewh-ebi:
Yep, it's a bit of extra client work, but I think default empty lists is cleaner. We can then avoid this messy thing in the EMPIAR ingest code: https://github.com/BioImage-Archive/ingest-from-empiar/blob/49a45d21c34281cc7f19184f2aa2446cfeb8a54a/empiar_ingest/utils.py#L232

PR #328, opened 2025-04-17
added label field to image, and updated field names for physical dimensions of image representations

sherwoodf:
Update to image representation fields to make it clear that we expect single voxel size and not the physical size captured by the whole field of view. This aligns it with the OME-ZARR data model, that only captures the information for the single pixel.

Additionally, especially for EMPIAR sumbission, adding an optional label field for images for some kind of human-reable title would be useful for the website.


liviuba (approved):
:+1:

PR #329, opened 2025-04-17
changes to attributes, provenance, and adding document creator

sherwoodf:
ticket: https://app.clickup.com/t/8698nhwmm

This PR aims to address our need to keep track of the source of data. Now that we have multiple ingestion sources and processes it is much less clear what the source of an object was.  We currently use AttributeProvenace to track the provenance of Attribute objects inside other objects. This extends the use of AttributeProvenance to a general Provenance enum for use on objects as well so we can capture source of objects. 

The change also includes renaming the attribute field to additional_metadata - which is somewhat more descriptive for such a general field - but importantly less confusing since attribute is an already overloaded term. 

Some areas which aren't addressed in the PR (yet) and would be good to encode in comments on the Provenance enum / would mean changes to the enum:
1. Do we want additional values to track whether something was e.g. ingested from EMPIAR vs ingested from Biostudies, or the output of image assignment from an empiar entry.
2. Additional clarity on when one value should be used vs another - should this just be based on the name of our code packages (including what source they're designed to process if it's not mentioned?)

I've not gone with making the submitter the default value (though the mock objects use it since they need _some_ value) because i suspect it won't be the default for most objects. From a discussion on Wednesday, it seems as though we may store 'what gets sent to us' separately (i.e. what's in biostudies, empiar, and in the future something else for maybe ro-crate objects?). So technically none of the objects would use the submitter value, as it would always be some part of our pipeline creating the object (and that is what we want to track). But it would still be useful to have a submitter value for tracking annotations data (in that here's some extra data written & structured by the submitter, not us)

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/attribute_models.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/attribute_models.py
@@ -1,10 +1,10 @@
 from __future__ import annotations
 
-from typing import List, Optional, Any
+from typing import Optional, Any
 from typing_extensions import Self
 
-from pydantic import BaseModel, Field, model_validator, field_validator, ValidationError
-from .semantic_models import Attribute, AttributeProvenance
+from pydantic import BaseModel, Field, model_validator, field_validator
+from .semantic_models import Attribute, Provenance
 from uuid import UUID
 
 # For shared models that are placed inside Attribute objects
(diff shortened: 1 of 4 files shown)


matthewh-ebi:
For 1. I don't think we should track whether 

For 2.

matthewh-ebi:
Happy with changes, thoughts on the questions:

1. I don't think we need to track EMPIAR/BioStudies ingest separately, seems cleaner not too (and avoids ending up with ro_crate_ingest or something similar). Similarly image assignment.
2. Code package by default for most of these seems sensible. This leaves `submitter` as being a bit unusual because we end up only using it for additional_metadata that's on File References. Not necessarily a problem.
 
I guess related to this - if we were taking RO-Crate submissions and loading them into the API, would we use `submitter` or `bia_ingest` for created objects?

sherwoodf:
> I guess related to this - if we were taking RO-Crate submissions and loading them into the API, would we use submitter or bia_ingest for created objects?

I can't think of a situation where ingest =/= 'submitter' in these situations, so i suspect we should just use one, and ingest seems the most consistent with the others. In ro-crate submissions, ingest might be the creator of images that we've not assigned, but were assigned by the user writing their ro-crate. But they could also get 'assigned' by the biostudies -> ro-crate process & then ingested....

PR #330, opened 2025-04-17
update title_id field name to title

sherwoodf:
Updating title_id field to just title

The original name comes from the fact that certain objects in biostudies are referenced by their title in order to connect them to e.g. associations. However:
- Once the object is in the api, uuids are used to reference the object - the title_id only gets used as a title
- We now ingest data from other sources than biostudies, where the title is just a title, and not an id

It was also sort of being used as a way of indicating which objects a user was editing (in that it was a "user id'd object") - again this isn't the case with non-biostudies submission, and we hope to cover modelling this with this other PR: https://github.com/BioImage-Archive/bia-integrator/pull/329


liviuba (approved):
:+1:

PR #331, opened 2025-04-17
added channel label to signal-channel-information

sherwoodf:
Added channel label to signal channel info to allow a mapping from image data to dyes & signal frequency information - particularly useful when an image is composed of multiple channels, each from a different dye & imaging at different illumination frequencies.

ticket: https://app.clickup.com/t/8698nhwmm


liviuba (approved):
:+1:

PR #332, opened 2025-04-28
wip individual arguments for elastic instead of connstring

liviuba:


api/helm-chart/templates/deployment.yaml
```diff
@@ -60,6 +60,11 @@ spec:
                 secretKeyRef:
                   name: {{ include "api.fullname" . }}
                   key: mongoConnstring
+            - name: ELASTIC_CONNSTRING
```
sherwoodf, line 63: Will the .env_compose and .env_template (and .env) need updating as well to not use constring in the future?

sherwoodf (approved):
Works for me locally - just have a small question about the deployment.yaml change

api/helm-chart/templates/deployment.yaml
liviuba: no, the container is still configured with the connstring, it's just the helm chart that needed the fields split (and then that builds the connstring to configure the api)

PR #333, opened 2025-04-29
RO-crate to integrator package refactor (part 1)

sherwoodf:
Added ro-crate model and example ro-crates, as well as various linked-data models to bia-shared-datamodels

No logic has changed from current ro-crate logic up to import statements / file path loading.


liviuba (approved):
:+1: 

Double-checking my understanding of the change, the underlying motivation for this was to be able to do the ro-crate -> api import?

sherwoodf:
Yep! And also eventually the biostudies -> rocrate!

PR #335, opened 2025-04-29
created uuid unique input attribute

sherwoodf:
Created sub-attribute to be used in any code that needs to create objects, in order to store the context of how the uuid was generated in the object sent to the api.

Only needed once uuid generation code has been merged from this branch.

### DIFF
--- a/bia-shared-datamodels/src/bia_shared_datamodels/attribute_models.py
+++ b/bia-shared-datamodels/src/bia_shared_datamodels/attribute_models.py
@@ -107,3 +107,30 @@ class DatasetAssociatedUUIDAttribute(Attribute, SubAttributeMixin):
         return self
 
     value: dict[str, list[str]] = Field()
+
+
+class DocumentUUIDUinqueInputAttribute(Attribute, SubAttributeMixin):
+    """
+    Model for storing the string that was used to generate the uuid of the object.
+    Note this does not need to contain the type nor the study uuid, since those should be findable in the rest of the document / through object links.
+    """
+
+    @field_validator("name", mode="after")
+    @classmethod
+    def validate_attribute_name(cls, value) -> Self:
+        if value != "uuid_unique_input":
+            raise ValueError(
+                f"Name for this type of attribute must be 'uuid_unique_input'"
+            )
+        return value
+
+    @field_validator("value", mode="after")
+    @classmethod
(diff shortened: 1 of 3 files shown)


bia-shared-datamodels/test/test_attribute_models.py
```diff
@@ -49,7 +53,7 @@ def test_sub_attribute_models(
             sub_attribute_model
         )
 
-        # Check basic attribute doesn't pass validation 
+        # Check basic attribute doesn't pass validation
```
kbab, line 56: Could this be put in a `with pytest.raises(ValidationError):` block?

bia-shared-datamodels/src/bia_shared_datamodels/attribute_models.py
```diff
@@ -107,3 +107,30 @@ def validate_attribute_value_key(self) -> Self:
...
+
+class DocumentUUIDUinqueInputAttribute(Attribute, SubAttributeMixin):
+    """
+    Model for storing the string that was used to generate the uuid of the object.
+    Note this does not need to contain the type nor the study uuid, since those should be findable in the rest of the document / through object links.
+    """
+
+    @field_validator("name", mode="after")
+    @classmethod
+    def validate_attribute_name(cls, value) -> Self:
```
kbab, line 120: is the use of `value` as opposed to `name` for the function parameter on purpose?

kbab (commented):
LGTM, happy to approve as is, but left some comments. Let me know if to just go ahead with approval. Also (not changed in this PR) line 513 `get_dataset_associatation_attribute` has a typo ...

bia-shared-datamodels/src/bia_shared_datamodels/attribute_models.py
sherwoodf: Yes, it was, but  a) i totally didn't realise why it would be confusing at the time and b) thought i had to to match the function structure.

All the pydantic examples: https://docs.pydantic.dev/latest/concepts/validators/#field-validators use value everywhere for after validators - which is why i used it here & for all the validators above.

... But i see now it's confusing in this case because we have a field called value, despite it not actually being related to why i called the field 'value' in the first place!!

I've just checked and changing it to name for this function still works - i'd want all of the validators to be consistent - do you have an preference between following the docs or having more specific naming here?

PR #339, opened 2025-04-30
deleted old models and regenerated client

sherwoodf:
Ran `make -C .. client.generate  ` after deleting all the files from the models folder in the client package.

Pre-requisite to various tickets for getting packages working with the new models.


liviuba (approved):
:+1: We might revisit the client generation at some point

PR #343, opened 2025-05-02
added api client saving and updated local save to follow the same pattern as other ingests

sherwoodf:
ticket: https://app.clickup.com/t/8698uqvab

Added logic to save api documents either as individual files, to a local api, or to the actual api, using env file to configure settings.

Added tests for both local and api saving.


ro-crate-ingest/ro_crate_ingest/save_utils.py
```diff
@@ -0,0 +1,152 @@
...
+        save_local_api(object_type, objects_to_save)
+    elif persistance_mode == PersistanceMode.BIA_API:
+        save_bia_api(object_type, objects_to_save)
+    else:
+        raise ValueError(
+            f"Something went wrong with the persistance mode: {persistance_mode}"
+        )
+
+
+def round_trip_object_class_from_client_to_datamodel(api_object):
```
liviuba, line 135: The assumption for idempotent writes was that the client writes anyway, and handles the failure (if any, ex with the version here).

With this function, it looks like we're reimplementing the write rejection but in a different place, is this intentional?

ro-crate-ingest/test/test_ro_crate_to_bia.py
```diff
@@ -1,12 +1,60 @@
...
+from ro_crate_ingest.cli import ro_crate_ingest
 import json
+from bia_integrator_api import models
+from pydantic.alias_generators import to_pascal
 
 runner = CliRunner()
 
 
-def test_ingest_ro_crate_metadata(tmp_path: Path):
+def test_ingest_ro_crate_metadata(tmp_bia_data_dir: Path):
```
liviuba, line 11: related to the question about supporting 3 writing modes, is writing to disk a core feature to have it tested separately?

liviuba (commented):
Left some comments, Looking at the PR title, the writing to disk and reimplementing some of the write logic around default values seems out of scope

ro-crate-ingest/ro_crate_ingest/save_utils.py
```diff
@@ -0,0 +1,152 @@
...
+
+logger = logging.getLogger("__main__." + __name__)
+
+
+class PersistanceMode(str, Enum):
+    """
+    Enum for the different persistance modes.
+    """
+
+    LOCAL_FILE = "local_file"
```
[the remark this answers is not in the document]
sherwoodf: Having write to local files is still useful for for a) easily checking whether the correct things are being written before sending to api and probably b) eventual tooling around preview of the 'end state' of submission. In either case, i viewed it more as  me 'updating the existing implementation that writes to disc to write things structured like the current ingest' rather than me 'implementing writing to disk'

In terms of local vs bia api - i think that's again helpful for testing?

ro-crate-ingest/test/test_ro_crate_to_bia.py
sherwoodf: It follows different enough code that i think i'd rather have it tested - but i guess that what i'd really want to know is that the things that get written locally are equivalent enough to what gets written in the api such that they can be used to verify what someone is sending to the api before running ingest.

ro-crate-ingest/ro_crate_ingest/save_utils.py
sherwoodf: So i was aware of bia ingest handling this - but i think it maybe that's because it was prior to getting idempotent writes working fully in the new api. My intent wasn't actually to replicate that feature, though as commented below, i think it is useful to have the local write to disk create things that look enough like what gets saved that it's a useful document to check.

Because what we send =/= what we get back from api - it makes it a bit annoying for testing. I guess maybe i can change the tests to be 'for every field in an object we send (if present), check that it matches what we get back'

liviuba (approved):
:+1: code looks correct

PR #344, opened 2025-05-06
remove use_type and object version bump

sherwoodf:
2 model changes:

1. remove use type from image rep models - this is because this information will be stored using attribute connections between the image and various image representations.
2. bump model version on all objects due to the addition of required fields - all objects have a new required field, and many objects have gone through somewhat significant change. Previous data would not be compatible, so a model version bump is sensible to signify the change.

Original ticket: https://app.clickup.com/t/8698nhx0n
And ticket i cut just to not forget about the change over the long weekend: https://app.clickup.com/t/8698xywp8


liviuba (approved):
Code looks correct

PR #347, opened 2025-05-08
Make assign image work with new models

kbab:
As a pre-requisite to migrating images and image representations we need to be able to assign images using 2025/04 models. This PR implements that. Clickup ticket: [https://app.clickup.com/t/8698zwuyj](https://app.clickup.com/t/8698zwuyj)


bia-assign-image/bia_assign_image/cli.py
```diff
@@ -110,7 +111,15 @@ def assign(
         bia_specimen = None
     else:
         bia_specimen = specimen.get_specimen(
-            image_uuid, image_preparation_protocol_uuid, bio_sample_uuid
+            study_uuid,
+            image_uuid,
+            # TODO: Discuss if these should be saved to dataset additional_metadata as lists
+            [
+                image_preparation_protocol_uuid,
```
sherwoodf, line 118: I'm surprised these aren't already lists

bia-assign-image/bia_assign_image/cli.py
```diff
@@ -75,17 +74,19 @@ def assign(
     assert len(set(dataset_uuids)) == 1
     submission_dataset_uuid = dataset_uuids[0]
     dataset = api_client.get_dataset(submission_dataset_uuid)
+    study_uuid = dataset.submitted_in_study_uuid
 
-    image_uuid = uuid_creation.create_image_uuid(file_reference_uuid_list)
+    image_uuid_unique_string = "".join([str(f) for f in file_reference_uuid_list])
```
sherwoodf, line 79: This is different to the function in l37 in image.py ("" vs " ")

So only gets defined once, I'd recommend either:
 a) this unique string be passed into the image.get_image function 
 b) there's a single uuid_creation.py file in bia_assign image that e.g. takes in a list of appropriate uuids and does the function

bia-assign-image/bia_assign_image/image_representation.py
```diff
@@ -24,64 +30,73 @@ def get_image_representation(
...
+    ## Copy file_pattern from image if it exists (only for UPLOADED_BY_SUBMITTER rep)
+    # file_pattern = next(
+    #    (attr for attr in image.attribute if attr.name == "file_pattern"), None
+    # )
+    # if file_pattern:
+    #    attribute.append(file_pattern)
+
+    if object_creator == semantic_models.Provenance.bia_image_assignment:
+        unique_string = f"{image.uuid}"
+    elif object_creator == semantic_models.Provenance.bia_image_conversion:
```
sherwoodf, line 54: I'm guessing this won't happen 'in regular production' after we've updated all the existing image uuids?

sherwoodf (changes requested):

bia-assign-image/bia_assign_image/cli.py
kbab: Will fix in ingest code

bia-assign-image/bia_assign_image/image_representation.py
kbab: The current workflow is that what used to be the 'UPLOADED_BY_SUBMITTER' representation is always created on image assignment. So I put the object_creator as bia_image_assignment in this case . This would still happen in regular production if we follow the previous workflow

kbab: Ahh - when we are converting images we need to create the image representation for the converted image - the current workflow calls this function. To my mind in this case the object creator is bia_image_conversion ...

sherwoodf (approved):

PR #348, opened 2025-05-08
Move Elastic refresh script to bia-export

liviuba:
https://app.clickup.com/t/8698ach8x

And adds the wrapper Helm chart we need to be able to re-index as a commad in the infra repo


api/scripts/elastic_refresh.sh
```diff
@@ -3,15 +3,15 @@ set -ex
 set -o pipefail
```
sherwoodf, line 4: If the refresh script is being moved, should it be deleted here?

sherwoodf (approved):
lgtm, with comment about leaving the original script in the api.

api/scripts/elastic_refresh.sh
liviuba: Left it in as a utility / alternative to having an export json as a test fixture, note it's just the script, not the Job / Helm part of it, and also the Docker monoimage is now api-only

Might delete it if unused, but I'd keep it around for now

liviuba: ^ Updated the PR now to make the api image api-only

PR #353, opened 2025-05-29
Fix client test after model changes

liviuba:
https://app.clickup.com/t/86997t963


liviuba:
Deleted use_type. Resolved comment about int/float mismatch, will reopen if it proves significant later

PR #354, opened 2025-05-29
Update ro crate models

sherwoodf:
Ticket: https://app.clickup.com/t/8698uqx6j

Updated example ro-crate make sure it passes the 'Must' validation rules from ro-crate & added tests

Updated field names to be more consistent with ro-crate structure, rather than our own - we may want to consider going further (e.g. changing displayName to Name for Person)

Also updated so created objects work for new 2025_04 models


AybukeKY (commented):
Couple minor comments/questions.

AybukeKY (approved):

PR #356, opened 2025-05-30
Update export for model updates

liviuba:
https://app.clickup.com/t/86998b5na


sherwoodf (approved):
Approved changes, tests can be fixed to deal with updated expectations later. I can't see any other changes that would affect export.

PR #357, opened 2025-06-02
update contributor role to be a list

sherwoodf:
Contributor in biostudies could be a string with comma separated values. It would be better represented in our API with a list.

Note have not changed export tests.


kbab (approved):
LGTM

PR #369, opened 2025-06-11
Factor out search

liviuba:
https://app.clickup.com/t/8699d1zrb

Moves currently approved search api as an independent subproject.


sherwoodf (approved):
LGTM in general - had to delete images to get the new compose/makefile to work, but then tests passed. I'm approving, but left a couple of nits in terms of Readme and folder naming.

PR #371, opened 2025-06-17
Add search by path name

Mandloi2309:
https://app.clickup.com/t/8699f4tc5

Task: Add a method to the API that allows an API consumer to search for a FileReference using its exact path_name.

Background: to correctly create links between images in our database (via the API that controls interaction with Mongo), we want to be able to set a particular attribute on the file(s) that comprise that image. We have old curation code that enables this, but to make it work with the more recent version of the API, we'll need to add an API method that allows searching for a FileReference object using one of its attributes (path_name).

### DIFF
--- a/api/api/search.py
+++ b/api/api/search.py
@@ -48,7 +48,7 @@ async def searchFileReferenceByPathName(
     path_name: Annotated[str, Query(min_length=5, max_length=1000)],
     db: Annotated[Repository, Depends(get_db)],
     pagination: Annotated[Pagination, Depends()],
-    study_uuid: shared_data_models.UUID = None,
+    study_uuid: shared_data_models.UUID,
 ) -> List:
 
     study = await db.get_doc(study_uuid, shared_data_models.Study)
--- a/api/api/search.py
+++ b/api/api/search.py
@@ -47,7 +47,6 @@ async def searchImageRepresentationByFileUri(
 async def searchFileReferenceByPathName(
     path_name: Annotated[str, Query(min_length=5, max_length=1000)],
     db: Annotated[Repository, Depends(get_db)],
-    pagination: Annotated[Pagination, Depends()],
     study_uuid: shared_data_models.UUID,
 ) -> List[shared_data_models.FileReference]:
     study = await db.get_doc(study_uuid, shared_data_models.Study)
@@ -88,7 +87,7 @@ async def searchFileReferenceByPathName(
     ]
     results = await db.aggregate(pipeline)
     if results:
-        results = [x for res in results for x in res["files"]][: pagination.page_size]
+        results = [x for res in results for x in res["files"]]
     return results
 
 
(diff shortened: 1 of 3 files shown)


liviuba (approved):
:+1:

api/api/search.py
```diff
@@ -43,6 +43,95 @@ async def searchImageRepresentationByFileUri(
     )
 
 
+@router.get("/file_reference/by_path_name")
+async def searchFileReferenceByPathName(
+    path_name: Annotated[str, Query(min_length=5, max_length=1000)],
+    db: Annotated[Repository, Depends(get_db)],
+    pagination: Annotated[Pagination, Depends()],
+    uuid: shared_data_models.UUID = None,
```
liviuba, line 51: just to make sure this makes its way to the client, what should it be a uuid for?

api/api/search.py
```diff
@@ -43,6 +43,95 @@ async def searchImageRepresentationByFileUri(
...
+) -> List:
+
+    if not uuid:
```
liviuba, line 54: this will be reasonably fast now because of the index, but if it's always going to be used for a specific study, we can make it non-optional for simplicity

api/api/search.py
```diff
@@ -43,6 +43,95 @@ async def searchImageRepresentationByFileUri(
...
+            {"file_path": path_name},
+            shared_data_models.FileReference,
+            pagination=pagination,
+        )
+
+    study = await db.get_doc(uuid, shared_data_models.Study)
+
+    results = await db.aggregate(
+        [
+            {"$match": {"uuid": uuid, "model.type_name": "Study", "model.version": 3}},
```
liviuba, line 65: I would reuse study.model.version here instead of a fixed value, so the aggregation would still work when we update the models

api/api/tests/test_search_file_reference.py
```diff
@@ -0,0 +1,87 @@
...
+        new_ref = existing_file_reference.copy()
+        new_ref["file_path"] = "test file path"
+        new_ref["uuid"] = get_uuid()
+        new_ref["submission_dataset_uuid"] = existing_dataset["uuid"]
+        rsp = api_client.post("private/file_reference", json=new_ref)
+        assert rsp.status_code == 201, rsp.json()
+        file_references.append(new_ref)
+
+    file_references.sort(key=lambda fr: fr["uuid"])
+    return {"file_references": file_references, "study_id": study_id}
```
liviuba, line 26: we can just return the list here, and if we request the study as a fixture in any test we get the same study that was used for the dataset

api/api/tests/test_search_file_reference.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_file_reference_path_name(api_client: TestClient, file_references_many: dict):
```
liviuba, line 29: on my local, this test passes if I run it with a clean db, but then if I re-run it it fails, not sure why

liviuba (commented):
Left some comments, especially the one around test_file_reference_path_name passing/not passing depending on what's in the db

api/api/search.py
Mandloi2309: Yes, it can be confusing. Will rename it to study_uuid.

PR #372, opened 2025-06-18
Example + ontology changes for ro-crates with filelists

sherwoodf:
updated ontology + ro-crate models for file lists and create new example ro-crate using file lists, using https://www.w3.org/ns/csvwto describe the tables.

Also noticed some things that weren't up to date in the ontology, namely, Association class was still being used despite the move to Association propeties directly on a dataset.

### DIFF
--- a/.github/workflows/core.yaml
+++ b/.github/workflows/core.yaml
@@ -52,7 +52,6 @@ jobs:
             bia-ingest,
             bia-assign-image,
             bia-converter,
-            ro-crate-ingest,
             bia-search
           ]
     runs-on: ${{ matrix.os }}


.github/workflows/core.yaml
```diff
@@ -52,7 +52,6 @@ jobs:
             bia-ingest,
             bia-assign-image,
             bia-converter,
-            ro-crate-ingest,
```
sherwoodf, line 55: this is intended - it's included in test below. It was accidentally added back in someone else's rebase.

Chr1st0p43rR (approved):
Hi Francois, all of this looks good to me — looking forward to part two!

PR #373, opened 2025-06-19
Update model diagram

sherwoodf:
Updated diagram to:
https://docs.google.com/drawings/d/1iWV7zQrCGiAK0bFm4gV4OVAcZ0VCXMRYiWiFAa8iISQ/edit
post model changes


matthewh-ebi (changes requested):
Small change/fix identified. I didn't go through each class to compare against the code to check full accuracy, think it might be easier to fix errors (if any) with future PRs than do this now.

matthewh-ebi (approved):
Looks good now!

PR #375, opened 2025-06-23
updated annotation method logic to handle list like annotation types

sherwoodf:
Minor update to deal with studies like S-BIAD1377 and others where the method type attribute is repeately set (rather than using a comma separate string)

As part of: https://app.clickup.com/t/8699g3tg7


tzulueta (approved):
looks good to me!

PR #378, opened 2025-06-27
Elastic api

liviuba:
https://app.clickup.com/t/8698znbec

Adds the ts client with website/* and search/*


clients/search/typescript/apis/DefaultApi.ts
```diff
@@ -0,0 +1,71 @@
+/* tslint:disable */
```
sherwoodf, line 1: It slightly looks like the files under apis/ and models/ are like a previous run at generating the client, and then they were generated under src/apis/ etc. with the new website/ endpoint. Are these duplicate or are they something else?

liviuba: deleted & regenerated the client, there actually were some stray files. Not adding delete/refresh to the script yet though, not sure that's what we want

Makefile
```diff
@@ -12,6 +12,10 @@ client.generate:
 client.examples:
 	docker compose --profile client_examples up --build --force-recreate --remove-orphans --abort-on-container-exit
 
+client.search.generate:
```
sherwoodf, line 15: i think this needs to be added to the phony

sherwoodf (approved):
LGTM!

PR #381, opened 2025-07-07
biostudies to ro-crate conversion

sherwoodf:
Ticket: https://app.clickup.com/t/8699atbv2

This is probably part 1 of many for this ticket.

Only covers v4 submissions. Does not attempt to cover file lists. 

I have moved some of the files & logic for the ro-crate -> bia api ingest under folders to make importing the relevant files easier, and have more separation between the files of the two sides of the pipeline.

Added some validation to the RO-Crate models (checking what goes into the @type field) after i made mistakes creating ro-crate objects in this PR, which i thought would be generally useful in other parts of the code base.

Added 2 end-to-end tests, focused on checking some of the more complex logic around creating the correct authors + affiliations, and the other around creating the correct biosamples, taxons, and getting the dataset associations to those correct.

### DIFF
--- a/ro-crate-ingest/Readme.md
+++ b/ro-crate-ingest/Readme.md
@@ -4,7 +4,7 @@ A CLI tool to convert between BioImage Archive API objects and JSON objects rela
 
 ## Install
 ```
-poetry env use python3.10
+poetry env use python3.13
 poetry install
 ```
 
@@ -23,6 +23,18 @@ This will create files for objects in the default cache location: ~/.cache/ro-cr
 
 the -p option can be used to save objects in a local api or the bia api. Settings for the BIA api can be set up by copying the .env_template file to an .env file, and filling in the details for your BIA account.
 
+
+## Biostudies to ro-crate
+
+Creates an ro-crate with an ro-crate-metadata.json and filelist per dataset:
+
+Example use:
+
+```
+poetry run ro-crate-ingest biostudies-to-roc S-BIAD843
+```
+
+Can choose the output folder with -c 
 
 ## Testing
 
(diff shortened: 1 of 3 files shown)


liviuba (approved):
:+1: 

A few notes:
1. I think the Readme needs the python version bumped
2.  I was having an issue with an attribute being `list[str]` but the input being `list[list[str]]` (error below)
3. Was surprised that this worked, but is there a reason for `convert` being named the same for both `ingest` and `biostudies-to-roc`?
4. Random q, but why are some types in the output ro-crate json namespaced and others not?

LE: Numbered list for referencing

sherwoodf:
> 1. I think the Readme needs the python version bumped

Will fix!

> 2. I was having an issue with an attribute being `list[str]` but the input being `list[list[str]]` (error below). Also in the ImagingMethod I think, there was a list default vs a default factory. Might cause issues if we re-run it

Thanks i will check that out

> 3. Was surprised that this worked, but is there a reason for `convert` being named the same for both `ingest` and `biostudies-to-roc`?

No reason - just copy-paste-and-forget-to-change - i too am surprised it worked - i guess because it's tagged differently in the cli? will update to be more consistent with commands

> 4. Random q, but why are some types in the output ro-crate json namespaced and others not?

This is because some stuff is in order to abide by the ro-crate-schema, and some stuff is for own own use. So, for instance, multiple things of type Dataset (https://schema.org/Dataset) because they correspond to a folder in the ro-crate schema, but might also be a bia:Study (if it's the root object with all the author & licence information), an bia:Dataset (if it's associating biosamples etc with a set of files), or even a bia:Image (if one image is made up of multiple files under the same folder, for timesteps, channels, etc)

PR #382, opened 2025-07-10
Propose source and annotation image

kbab:
PR for [https://app.clickup.com/t/8699g3mq1](https://app.clickup.com/t/8699g3mq1) . Allows getting source images and related annotation images for processing.

### DIFF
--- a/bia-assign-image/bia_assign_image/cli.py
+++ b/bia-assign-image/bia_assign_image/cli.py
@@ -211,6 +211,47 @@ def propose_images(
         logger.info(f"Wrote {count} proposals for {accession_id} to {output_path}")
 
 
+@app.command(
+    help="Propose file references for source images and associated annotations to convert for an accession"
+)
+def propose_images_and_annotations(
+    accession_ids: Annotated[
+        List[str], typer.Argument(help="Accession IDs to process")
+    ],
+    output_path: Annotated[Path, typer.Argument(help="Path to write the proposals")],
+    api_target: Annotated[
+        ApiTarget, typer.Option("--api", "-a", case_sensitive=False)
+    ] = ApiTarget.prod,
+    max_items: Annotated[
+        int, typer.Option(help="Maximum number of items to propose")
+    ] = 5,
+    append: Annotated[
+        bool, typer.Option(help="Append to existing file instead of overwriting")
+    ] = True,
+    check_image_creation_prerequisites: Annotated[
+        bool,
+        typer.Option(
+            help="Check whether dataset linked to file reference contains requirements needed to create a bia_data_model Image object."
+        ),
+    ] = False,
+) -> None:
+    """Propose file references for source images and associated annotations to convert for an accession IDs"""
+    for accession_id in accession_ids:
+        n_source_images, n_annotations = (
+            propose.write_convertible_source_annotation_file_refs_for_acc_id(
+                accession_id,
+                output_path,
+                api_target=api_target,
+                max_items=max_items,
+                check_image_creation_prerequisites=check_image_creation_prerequisites,
+                append=append,
+            )
+        )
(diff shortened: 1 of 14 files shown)


bia-assign-image/README.md
```diff
@@ -42,6 +42,28 @@ poetry run bia-assign-image assign-from-proposal proposals.yaml
 
 This will create BIA Image objects and default representations for each proposed file reference.
 
+### Proposing Images and Associated Annotations
+In the case where a study has annotation datasets, and the file list column has a 'source_image' entry, the ingest process creates a 'source_image_uuid' entry in the additional metadata of annotation file references. The `propose-images-and-annotations` command can be used to create a proposal file with source images and their respective annotation images. It is advisable to use this command with `--no-check-image-creation-prerequisites` in case the annotation datasets are missing information on associations to bio-samples, sample image preparation protocols etc.
```
sherwoodf, line 46: --no-check-image-creation-prerequisites is set to false by default, but also, the expectation is that annotation datasets will be missing information to associations to bio-samples etc, so i'm not sure that ever making these checks makes sense? It would probably be better to check in this case that the associated dataset had an annotation method.

bia-assign-image/bia_assign_image/propose.py
```diff
@@ -204,6 +184,85 @@ def write_convertible_file_references_for_accession_id(
...
+
+    convertible_file_references = get_convertible_file_references(
+        accession_id,
+        api_target,
+        check_image_creation_prerequisites,
+        include_annotation_details=True,
+    )
+
+    # Below assumes ingest has populated 'source_image_uuid'
+    # Note: source_image_uuid is uuid of file reference of raw image - NOTHING to do with bia_shared_datamodel.Image objects!)
```
sherwoodf, line 221: That's a good point - i named it that since it's based on the 'source image' column from the file reference, but maybe this is overly confusing. Since it's an attribute it would be pretty easy to change on the objects with a re-ingest.

bia-assign-image/bia_assign_image/cli.py
```diff
@@ -211,6 +211,47 @@ def propose_images(
         logger.info(f"Wrote {count} proposals for {accession_id} to {output_path}")
 
 
+@app.command(
```
sherwoodf, line 214: Not related to this pr, but --help doesn't work (even on main)

bia-assign-image/bia_assign_image/propose.py
```diff
@@ -204,6 +184,85 @@ def write_convertible_file_references_for_accession_id(
...
+    # index annotation images by their corresponding source image uuids
+    index = {}
+    for fr in convertible_file_references.values():
+        source_image_uuid = fr.get("source_image_uuid")
```
sherwoodf, line 225: if it's clearer, i would be happy with you re-naming the variable here to something like source_image_file_ref_id

bia-assign-image/bia_assign_image/propose.py
```diff
@@ -298,3 +357,39 @@ def read_proposals_from_yaml(proposal_path: Path) -> List[Dict]:
...
+    file_references: list[bia_data_model.FileReference],
+    include_annotation_details,
+) -> dict[str, dict]:
+    file_reference_details = {}
+
+    for fr in file_references:
+        if not in_bioformats_single_file_formats_list(fr.file_path):
+            continue
+
+        fr_details = {
```
sherwoodf, line 375: this could be a class, since it gets passed around a fair bit - it would make the type hints using this clearer than expecting a dict.

sherwoodf (approved):
Approved with stylistic suggestions only. Validation (check image creation prerequisitis0 is probably the only thing worth having a followup ticket for perhaps?

PR #385, opened 2025-07-17
Using new uuid functions - ingest

sherwoodf:
update ingest to use new functions plus created function for default biostudies template

Should be a 'no-op' change


Chr1st0p43rR (approved):
All looks good to me.

PR #395, opened 2025-07-22
Fix image additional metadata

kbab:
Make migration script use latest uri format. [Clickup ticket](https://app.clickup.com/t/8699v8bf3)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,8 @@
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
     "python.analysis.extraPaths": [
-        "./clients/python"
+        "./clients/python",
+        "./bia-assign-image/scripts"
     ],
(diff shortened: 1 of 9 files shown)


bia-assign-image/scripts/fix_additional_metadata.py
Chr1st0p43rR, line 1: Just so I'm clear on this — these functions are not called by anything (besides tests), and that's supposed to be the case, right? Given the TODO at the bottom.

.vscode/settings.json
Chr1st0p43rR, line 1: A nitpick, but is it worth making a settings.json within bia-assign-image? For that it keeps things maybe more organised, and vs code tells me it can't find `fix_additional_metadata` in the tests (that is, opening just bia-assign-image in vs code — a "local" settings.json means the scripts can be found).

Chr1st0p43rR (approved):
Looks good, with one clarification and one nitpick.

PR #406, opened 2025-08-07
Add facets

liviuba:
Ticket: https://app.clickup.com/t/8699zr4tj


Mandloi2309 (approved):
LGTM, I could filter on the fts using organism, year and imaging method with lists as well. 
Only warning i got during running the test was 
`Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x1087f27b0>
Unclosed connector`
i am not sure if this needs to be handled or not apart from this all looks good!

PR #412, opened 2025-08-15
Add image index

liviuba:
https://app.clickup.com/t/8699zwd3d

Adds images to the search api, with just the uuid indexed now
Adds the same facet parameter naming to https://alpha.bioimagearchive.org/search as the one used by the client-side front-end (facet.*)


Mandloi2309 (approved):
I was trying to find images other than `445cb17b-95dc-47a3-9efc-b7a1a066bb51` one in the test, but could not. Does it have to do indexes? Apart from it looks good!

liviuba:
Which uuids did you try? I also looked up uuids from https://alpha.bioimagearchive.org at first and couldn't find any images since the website is still static.

But it looks like all exported images are now in search, ex https://alpha.bioimagearchive.org/search/search/fts/image?query=1e02db9b-d273-4562-a95e-e78d78c26fd7

PR #414, opened 2025-08-19
Handle empiar starfiles as annotation data

sherwoodf:
Ticket: https://app.clickup.com/t/869a4a328

Adds handling of star files from empiar submission as annotation data. This creates all relevant objects in the same manner as the existing empiar ingest, except for 1 thing: the attribute on the annotation data objects which label which part of the annotation data file correspond to the file is missing. I'm not aware of us using this for anything yet, so i think this passes the requirements of the ticket. I can either do a quick solution to this (which involves generating curation yamls to run curation on the object) but would rather solve this in a large ticket to handle fields that aren't covered by our own fields in general.

Note: does not handle annotation data in ro-crate metadata, only in file lists. I have cut another ticket: https://app.clickup.com/t/869a685dn in which i want to test for & implement the generic case of images being named / identified in both a file list and an ro-crate at the same time. This isn't what the empair or biostudies ingest currently does, but we need to have clearly tested & defined behaviour for it.

### DIFF
--- a/ro-crate-ingest/ro_crate_ingest/empiar_to_ro_crate/empiar/entry_api.py
+++ b/ro-crate-ingest/ro_crate_ingest/empiar_to_ro_crate/empiar/entry_api.py
@@ -11,9 +11,8 @@ logger = logging.getLogger("__main__." + __name__)
 def load_empiar_entry(accession_id) -> Entry:
     # Note this is a dictionary to include reasons why the override was made
     overrides = {
-        "EMPIAR-ANNOTATIONTEST": "A test submission.",
-        "EMPIAR-IMAGEPATTERNTEST": "A test submission.",
-        "EMPIAR-STARFILETEST": "A test submission.",
+        "EMPIAR-IMAGEPATTERNTEST": "A test submission covering file pattern shapes.",
+        "EMPIAR-STARFILETEST": "A test submission covering star file annotation, with tomograms, and other image dependencies.",
     }
     if accession_id in overrides:
         return read_empiar_entry_override(accession_id)
--- a/ro-crate-ingest/ro_crate_ingest/empiar_to_ro_crate/empiar/file_api.py
+++ b/ro-crate-ingest/ro_crate_ingest/empiar_to_ro_crate/empiar/file_api.py
@@ -14,9 +14,8 @@ class EMPIARFile(BaseModel, frozen=True):
 def get_files(accession_id: str) -> list[EMPIARFile]:
     # Note this is a dictionary to include reasons why the override was made
     overrides = {
-        "EMPIAR-ANNOTATIONTEST": "A test submission.",
-        "EMPIAR-IMAGEPATTERNTEST": "A test submission.",
-        "EMPIAR-STARFILETEST": "A test submission.",
+        "EMPIAR-IMAGEPATTERNTEST": "A test submission covering file pattern shapes.",
+        "EMPIAR-STARFILETEST": "A test submission covering star file annotation, with tomograms, and other image dependencies.",
     }
     if accession_id in overrides:
         return get_file_info_from_override(accession_id)


ro-crate-ingest/ro_crate_ingest/empiar_to_ro_crate/entity_conversion/file_list.py
```diff
@@ -132,36 +137,72 @@ def find_matching_imageset_and_images(
...
+            "type": bia_type,
+            "label": label,
+            "source_image_label": source_image_label,
+            "associated_annotation_method": associated_annotation_method,
+            "associated_protocol": associated_protocol,
         }
     )
 
 
 def split_dataframes_by_dataset(file_df: pd.DataFrame):
```
matthewh-ebi, line 194: docstring/return type would help understand what this function's doing

ro-crate-ingest/ro_crate_ingest/ro_crate_to_api/api_conversion.py
```diff
@@ -99,16 +99,16 @@ def convert_ro_crate_to_bia_api(
         get_settings().parallelisation_max_workers,
     )
 
-    if not image_raw_dataframe["image_id"].isna().all():
+    if not image_raw_dataframe["result_data_id"].isna().all():
```
matthewh-ebi, line 102: Comment to help understanding of this row would be good - without pandas background knowledge it's hard to parse

ro-crate-ingest/ro_crate_ingest/ro_crate_to_api/entity_conversion/file_reference.py
```diff
@@ -150,23 +151,41 @@ def select_size_in_bytes(row: dict) -> int:
     raise ValueError(f"No dataset found for file: {row["path"]}")
 
 
-def select_image_id(row: dict, image_extensions: list[str]) -> Optional[str]:
+def select_result_data_id_and_type(row: dict, image_extensions: list[str]) -> Optional[str]:
```
matthewh-ebi, line 154: Return type wrong here? Looks to be returning a tuple (either of which is None?)

ro-crate-ingest/ro_crate_ingest/ro_crate_to_api/entity_conversion/file_reference.py
```diff
@@ -150,23 +151,41 @@ def select_size_in_bytes(row: dict) -> int:
...
+            bia_type = "http://bia/Image"
 
-    standardised_file_extension = get_standardised_extension(row["path"])
-    if standardised_file_extension in image_extensions:
-        return row["path"]
+    return id, bia_type
 
 
-def select_image_label(row: dict) -> Optional[str]:
+def select_result_data_label(row: dict) -> Optional[str]:
```
matthewh-ebi, line 188: Definitely could use docstring here, we use different types of ids in different bits of code and it would help to understand which we're dealing with here

ro-crate-ingest/ro_crate_ingest/ro_crate_to_api/entity_conversion/result_data_and_dependency_creation.py
```diff
@@ -283,3 +231,52 @@ def create_uuid_list(
...
+
+    persist(
+        accession_id,
+        APIModels.AnnotationData,
+        [APIModels.AnnotationData(**model_dict)],
+        persistence_mode,
+    )
+
+
+def caluclate_dependency_height(df):
```
matthewh-ebi, line 264: caluclate -> calculate, more critically some explanation of what we're doing here (graph traversal things?)

matthewh-ebi (changes requested):
Looks great overall, the requested changes are all readability related.

I don't want to hold up this PR, but for the complex logic we’ll be heavily dependent on in future, we might want some unit type tests to cover edge cases and failure modes. File list processing (particularly matching image / annotation labels for on), and probably the dependency graph processing code as well, that looks gnarly and full of pit traps.

ro-crate-ingest/ro_crate_ingest/empiar_to_ro_crate/entity_conversion/file_list.py
```diff
@@ -132,36 +137,72 @@ def find_matching_imageset_and_images(
...
+                pattern = annotation_data.get("file_pattern", None)
+                if pattern:
+                    result = parse.parse(f"data/{pattern}", path)
+                    if result is not None:
+                        label = annotation_data["label"]
+                        bia_type = "http://bia/AnnotationData"
+                        source_image_label = annotation_data.get(
+                            "input_image_label", None
+                        )
+                        if is_list_of_dicts(source_image_label):
```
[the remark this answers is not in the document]
sherwoodf: I've realised this is actually where i've changed the case the way the yaml should be formatted, so i can simply the logic. It's actually always yaml objects.

PR #418, opened 2025-08-21
Generate ro-crate from bsst submissions files in pagetab

sherwoodf:
added processing for default template bsst submissions where the files are placed in the pagetab

ticket: https://app.clickup.com/t/8699zncyr

Also removed some unused imports, and neatened up global variable setting, as it was causing issue when multiple tests were run.

### DIFF
--- a/ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/entity_conversion/contributor.py
+++ b/ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/entity_conversion/contributor.py
@@ -16,21 +16,21 @@ from bia_shared_datamodels import ro_crate_models
 logger = logging.getLogger("__main__." + __name__)
 
 
-
 def get_contributors(
     submission: Submission,
     roc_affiliation_by_accno: dict[str, ro_crate_models.Affiliaton],
 ) -> list[ro_crate_models.Contributor]:
 
-    global CONTRIBUTOR_BNODE_INT
-    CONTRIBUTOR_BNODE_INT = 0
-
+    id_counter = 0
     sections = find_sections_recursive(submission.section, ["author"])
 
     contributors = []
     for section in sections:
 
-        contributors.append(get_contributor(section, roc_affiliation_by_accno))
+        contributor, id_counter = get_contributor(
+            section, roc_affiliation_by_accno, id_counter
+        )
+        contributors.append(contributor)
 
     return contributors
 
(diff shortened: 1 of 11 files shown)


ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/entity_conversion/contributor.py
```diff
@@ -16,14 +16,15 @@
...
 
-CONTRIBUTOR_BNODE_INT = 0
-
 
 def get_contributors(
     submission: Submission,
     roc_affiliation_by_accno: dict[str, ro_crate_models.Affiliaton],
 ) -> list[ro_crate_models.Contributor]:
 
+    global CONTRIBUTOR_BNODE_INT
```
karatugo, line 25: I think the consensus is to avoid `global` unless absolutely necessary.

ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/entity_conversion/pagetab_file.py
```diff
@@ -0,0 +1,87 @@
...
+)
+from urllib.parse import quote
+from pathlib import Path
+from ro_crate_ingest.save_utils import write_filelist
+
+
+logger = logging.getLogger("__main__." + __name__)
+
+
+def get_dataset_and_filelist_for_pagetab_files(
```
karatugo, line 25: Probably because I'm not familiar with the naming, I found the name of the function and the return variable confusing. We return `file_list_and_dependencies` but the function is called `get_dataset_and_filelist`. Is dataset dependency of filelist?

ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/entity_conversion/pagetab_file.py
```diff
@@ -0,0 +1,87 @@
...
+    write_filelist(output_ro_crate_path, filelist_id, dataframe_filelist)
+
+    filelist = create_ro_crate_filelist_and_schema_objects(
+        filelist_id=filelist_id,
+        column_headers=dataframe_filelist.columns.values.tolist(),
+        column_by_name_url=column_by_name_url,
+        schema_list=schema_list,
+    )
+
+    column_list = []
```
karatugo, line 83: I think we can shorten this: 

```
column_list = [col for x in column_by_name_url.values() for col in x.values()]
```

karatugo (approved):
LGTM! Thanks!

ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/entity_conversion/contributor.py
sherwoodf: I need to pass this int between function to occasionally increment. Since an int isn't a mutable, python will create a copy of the int when i pass it to the function. So i can't just update the int in the function.

I could change the get_contributor function to return the int it uses as well as the object it create, and therefore update value in get_contributors. I'll see if that looks okay readability wise despite being an odd way to do an incremental count.

PR #421, opened 2025-08-29
Override for S-BIAD2250

Mandloi2309:
https://app.clickup.com/t/869a9zrcp

- Specimen title not matching with the relations


bia-ingest/bia_ingest/biostudies/api.py
```diff
@@ -245,6 +246,7 @@ def read_override(accession_id: str) -> Submission:
     )
     abs_path = submission_path.absolute()
     logger.info(f"Reading submission from {abs_path}")
+    logger.info(f"Absolute Path: {abs_path}.")
```
sherwoodf, line 249: I think we can remove this. Do you want to include the relative path fix?

sherwoodf (approved):

PR #423, opened 2025-09-02
include OIB for propose

Mandloi2309:
- OIB is supported by BioFormats


sherwoodf (approved):
Pictures looked good to me!

PR #424, opened 2025-09-04
propper handling of protocol and annotation method in EMPIAR-> RO crate conversion

sherwoodf:
fix empiar to ro-crate conversion handling of protocols and annotation methods into the csv and added handling of protocols in the ro-crate-metadata.json

Changed tests to use pytest_check when comparing output: this lets the tests check all the objects and return all the differences, which helps understand what's going on with test failures.

ticket: https://app.clickup.com/t/869aa1xq8


Chr1st0p43rR (approved):
This looks good to me, thanks!

PR #428, opened 2025-09-10
Add images refresh

liviuba:
https://app.clickup.com/t/869a4pabg (bugfix)

Adds images to the index & bugfix. Still TODO to have the index uniquely defined / less error-prone


liviuba:
Thanks! Depends on the size of the documents, the database, etc

PR #441, opened 2025-09-25
869af87eq add doi field to study model and api

Cesarec88:


bia-shared-datamodels/src/bia_shared_datamodels/utils.py
sherwoodf, line 1: In general, we try to provide a more specific/descriptive name than utils, but that's more of an issue when there's all sorts of different functions being thrown in there.

bia-shared-datamodels/src/bia_shared_datamodels/utils.py
```diff
@@ -0,0 +1,19 @@
+import re
+from urllib.parse import urlparse, unquote
+
+_DOI_RX = re.compile(r"^10\.\d{4,9}/[-._;()/:A-Z0-9]+$", re.I)
+
+
+def sanitise_doi(value: str) -> str:
+    s = value.strip()
```
sherwoodf, line 8: short variable names ;)

sherwoodf (approved):

PR #446, opened 2025-10-03
created valid jsonld context object

sherwoodf:
Making the example context a .jsonld file and nested the context inside a @context field


liviuba (approved):
Approving the actual change, with a few questions:

1. Where are we going to host this so that the relative context example applies?
2. Related to 1, do we have a preference for link header vs context attribute link vs context attribute object, for adding it to existing apps?

PR #546, opened 2026-04-01
Fix roc submission model exceptions

kbab:
Fix submission parsing issue due to some sections being in lists. This fix means no errors in parsing pagetab json. However, modified code to raise errors if we encounter sections with data that will not be processed. Clickup ticket: [https://app.clickup.com/t/869cn16a0](https://app.clickup.com/t/869cn16a0).

However, raising errors when sections can't be processed means, we either need to process all expected sections (e.g. Publication), or have a list of excepted sections the we are happy that we are not currentlly processing. Will discuss with @sherwoodf and create clickup ticket if necessary.

### DIFF
--- a/ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/biostudies/submission_api.py
+++ b/ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/biostudies/submission_api.py
@@ -20,15 +20,6 @@ FLIST_URI_TEMPLATE = (
 FILE_URI_TEMPLATE = "https://www.ebi.ac.uk/biostudies/files/{accession_id}/{relpath}"
 
 
-# def flatten_lists_of_list(data) -> list:
-#    if isinstance(data, list):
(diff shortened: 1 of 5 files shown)


ro-crate-ingest/ro_crate_ingest/biostudies_to_ro_crate/biostudies/submission_api.py
```diff
@@ -20,13 +20,16 @@
 FILE_URI_TEMPLATE = "https://www.ebi.ac.uk/biostudies/files/{accession_id}/{relpath}"
 
 
-def flatten_lists_of_list(data) -> list:
+def flatten_lists_of_list(data, parent_list=None) -> list:
     if isinstance(data, list):
```
sherwoodf: I think we should maybe return something or have an else that raises. Currently i think the function return type is actually -> list | None.

sherwoodf (changes requested):

sherwoodf (approved):
LGTM

PR #548, opened 2026-04-02
Handle datasets that have associations when no objects of those types were found

sherwoodf:
skip trying to associate datasets with certain types of object if no objects of that type have been created in the study

Move towards providing a unified interface over all the biostudies -> ro-crate mappers. I've not created mapping classes for all ro-crate classes, and i think that datasets would benefit the most of all the ones left for having one (as there are some nice class level variables to handle)

e.g. poetry run ro-crate-ingest biostudies-to-roc S-BIAD2283        

from https://docs.google.com/spreadsheets/d/1nMmNn5IkFnjhJcJkY-b3DiiJom8_IQAr0BvlMfeaXPU/edit?gid=0#gid=0


kbab (approved):
LGTM and all tests passed on my machine. However, I left a couple of comments about lines where I wasn't sure if the assignments could have issues.

PR #551, opened 2026-04-07
Fix minor bug in name of annotation method in dataset.py 

kbab:
Fixed minor bug in referring to annotation method in [PR 548](https://github.com/BioImage-Archive/bia-integrator/pull/548). The main issue is a one-line typo in `dataset.py`, but added tests for annotation method as it may be useful in future.


Chr1st0p43rR (approved):
Looks good to me, and checked tests run ok locally.

PR #552, opened 2026-04-14
Fix filelist type and biosample title bugs

kbab:
Original aim is to fix minor issue with biosample title and assignment of files in filelist to bia:Image (or bia:Annotation). However, encountered a bug due to multiple associations which is also addressed. Clickup Ticket is [here](https://app.clickup.com/t/869cn3gzu)

### DIFF
--- a/ro-crate-ingest/override_data/biostudies/S-BIADTEST_COMBINE_FILELIST/file_list_test_1.json
+++ b/ro-crate-ingest/override_data/biostudies/S-BIADTEST_COMBINE_FILELIST/file_list_test_1.json
@@ -1,6 +1,6 @@
 [
   {
-    "path": "folder1/file1.ome.zarr.zip",
+    "path": "folder1/file1.not.image.extension",
     "size": 24628037,
     "attributes": [
       {
--- a/ro-crate-ingest/override_data/biostudies/S-BIADTEST_COMBINE_FILELIST/file_list_test_2.json
+++ b/ro-crate-ingest/override_data/biostudies/S-BIADTEST_COMBINE_FILELIST/file_list_test_2.json
@@ -19,7 +19,7 @@
     "type": "file"
   },
   {
-    "path": "folder2/file2.ome.zarr.zip",
+    "path": "folder2/file2.not.image.extension",
     "size": 13161534,
     "attributes": [
       {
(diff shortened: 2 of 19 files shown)


sherwoodf:
I don't know that the filelist type column is a bug, nor that we necessarily want this fix.

The existing column is called type, but doesn't connect to the rdf:type property, and therefore isn't used to assign images, and is generally ignored (it will show up in the created file references). 

We used to create images in the ro-crate -> api portion of the processing based on image extension. We stopped because we didn't want to do this automatically. Doing so here is doing so at an earlier step, prior to review, but is still automatically performing this processing.... I'm not sure how much we want to do that. The approach typically assigned the maximal number of images that could be created (except for the case one file -> many images, which i suspect would mostly be just zips?). Since we typically try to avoid object deletion, i think we would want this to be tied into image assignment and ro-crate improvement pipelines that chris is working on.

sherwoodf (changes requested):

kbab:
> I don't know that the filelist type column is a bug, nor that we necessarily want this fix.

Thanks - your comment makes sense in the light of the approach to image assignment going forward. I was encountering the issue because I was trying to do the end-to-end process (i.e. biostudies-to-roc followed by roc-to-api) automatically.

I will revert the changes and leave the type column here empty


> The existing column is called type, but doesn't connect to the rdf:type property, and therefore isn't used to assign images, and is generally ignored (it will show up in the created file references).

the biostudies API returns a 'type' which is usually one of 'file' or 'directory'. Would this value be suitable for the 'type' column instead? In the past we used it to distinguish between things that were stored as files or as zips (directories get converted to zips).

sherwoodf:
> > The existing column is called type, but doesn't connect to the rdf:type property, and therefore isn't used to assign images, and is generally ignored (it will show up in the created file references).
> 
> the biostudies API returns a 'type' which is usually one of 'file' or 'directory'. Would this value be suitable for the 'type' column instead? In the past we used it to distinguish between things that were stored as files or as zips (directories get converted to zips).

No, it wouldn't be suitable, as the value should be an rdf class. I suspect we will have to rename this as biostudies_type as part of the creation of the appropriate column. That might be worth doing as part of this processing, since we willl probably want to do it for all filelists, and it will only affect the attribute field of the file references.

kbab:
Closed. Comments addressed in new [PR560](https://github.com/BioImage-Archive/bia-integrator/pull/560)

PR #554, opened 2026-04-15
Use bia ro crate

sherwoodf:
Changes:
- Remove bia_ro_crate from ro_crate_ingest - that is now copied over to it's own package
- Remove validator from ro_crate_ingest - that is copied over to bia_ro_crate package. CLI command remains pointing to new package's validator command.
- Remove validator tests from ro_crate_ingest - they are copied over to bia_ro_crate package
- Remove ro_crate_models from bia_shared_datamodels - that is now copied over to bia_ro_crate package
- Remove linked_data from bia_shared_datamodels - that is now copied over to bia_ro_crate package
- Remove ontology tests from bia_shared_datamodels - they are copied over to bia_ro_crate package
- Update ro_crate_ingest imports to use bia_ro_crate
- Update bia_shared_datamodels to remove unused rdf etc. imports
- Update context creation to point at new path in bia_ro_crate. I think this would be better as a function in that package, but i'm trying to limit the scope of changes here since the refactor PR is so bit already.


Chr1st0p43rR (approved):
Refactor looks good to me, and thanks for including the changes from #553.

PR #555, opened 2026-04-15
Update s3 url to livingobjects

kbab:
Update uri's and bucket names to reflect migration to livingobjects S3 storage. Clickup ticket [here](https://app.clickup.com/t/869cxhurg)


Chr1st0p43rR (approved):
Looks good, thanks. As we had discussed, I had difficulty initially, getting the tests running locally — I think there was some sort of falling back to a pytest from outside the venv, since it's not in the pyproject.toml, and it fails confusingly, with import errors, not _not_ finding pytest. And as you suggested, adding it did the trick, so we could always add:
```
[tool.poetry.group.dev.dependencies]
pytest = "^9.0.3"
```
to avoid that? But feel free to ignore that and go ahead and merge!

kbab:
Thanks Chris,

I tried the below in an earlier commit, but CI-CD was not happy. I have merged, and we can discuss in person on Tuesday 21/04.

> Looks good, thanks. As we had discussed, I had difficulty initially, getting the tests running locally — I think there was some sort of falling back to a pytest from outside the venv, since it's not in the pyproject.toml, and it fails confusingly, with import errors, not _not_ finding pytest. And as you suggested, adding it did the trick, so we could always add:
> 
> ```
> [tool.poetry.group.dev.dependencies]
> pytest = "^9.0.3"
> ```
> 
> to avoid that? But feel free to ignore that and go ahead and merge!

PR #556, opened 2026-04-20
updated ID for annotation components

sherwoodf:
Ticket: https://app.clickup.com/t/869cz5udj

Pre-pend a _ for annotation method IDs to make sure they don't clash with Dataset Ids
Update uuid function to still generate the same uuids

added to combine file list test to showcase overlapping file paths - need to work out how to fix in a future PR.


kbab (approved):
LGTM

PR #559, opened 2026-04-22
Empiar file list property

sherwoodf:
Update uuid functions, and add back in end-to-end tests


ro-crate-ingest/test/test_ro_crate_to_bia.py
```diff
@@ -46,6 +46,9 @@ def get_empiar_to_ro_crate_path(accession_id) -> Path:
     return Path(__file__).parent / "empiar_to_ro_crate" / "output_data" / accession_id
 
 
+def get_minimal_to_ro_crate_path(accession_id) -> Path:
+    return Path(__file__).parent / "minimal_to_ro_crate" / "output_data" / accession_id
```
Chr1st0p43rR: I realise that this doesn't match the path to the test minimal ro-crates, as it is — I must have misnamed the folder, leaving out the "to" — but currently doesn't affect the tests, because it's not used.

Chr1st0p43rR (approved):
Looks good, with one comment about the minimal ro-crate folder name in the tests (which I think would've been my mistake previously). Thanks!

PR #563, opened 2026-04-24
update test expectations after multiple prs were merged without rebasing

sherwoodf:
Prs added tests which needed updating after other PRs were merged - specifically those fixing biosample titles


kbab (approved):
LGTM - however, I was not sure why the url encoding codes were replaced with characters.

